PR #8, opened 2025-08-27
feat(py): Setup python bindings project

aborgna-q:
Setups a python project under `impl/py`, so we can publish the `jeff-format` package.

There are some problems with setting up dependencies, but we can merge this as-is for the moment.

The code from #3 is left unchanged apart from adding the capnp schema to a sub-dir and running automatic linters.


impl/py/src/jeff/capnp.py
```diff
@@ -0,0 +1,21 @@
+"""Utility functions for tket extensions."""
```
dime10, line 1: Docstring needs an update

impl/py/src/jeff/data/jeff.capnp
dime10, line 1: We should be able to include this file in the wheel via payload data, rather than duplicating the source

impl/py/src/jeff/__init__.py
```diff
@@ -397,7 +408,9 @@ def inputs(self) -> list[JeffValue]:
         if self._inputs is not _Empty:
             return self._inputs
 
-        return [JeffValue.from_encoding(inp, self._func) for inp in self._raw_data.inputs]
+        return [
```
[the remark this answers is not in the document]
aborgna-q: Sure 👍

PR #10, opened 2025-08-28
Add additional docs (FAQs and motivation)

josh146:


docs/faqs.md
```diff
@@ -0,0 +1,131 @@
...
+with QIR.
+
+### Why not HUGR?
+
+[HUGR](https://github.com/CQCL/hugr) is a hybrid quantum-classical program representation introduced by Quantinuum and used by the
+[TKET](https://github.com/CQCL/tket) and [Guppy](https://github.com/CQCL/guppylang) software libraries. While satisfying a lot of our intended criteria, HUGR itself was
+not designed to be an exchange format, and instead is used as an internal intermediate
+representation. In addition, HUGR contains advanced features such as type structure that has a high
+barrier to entry.
+
```
josh146, line 131: any other questions I am missing?

docs/faqs.md
```diff
@@ -0,0 +1,131 @@
...
+broadly.
+
+Finally, as per QIR the software tooling remains a big issue. Not all compilers are using LLVM,
+which MLIR depends upon, and it introduces the same significant barrier to entry as discussed above
+with QIR.
+
+### Why not HUGR?
+
+[HUGR](https://github.com/CQCL/hugr) is a hybrid quantum-classical program representation introduced by Quantinuum and used by the
+[TKET](https://github.com/CQCL/tket) and [Guppy](https://github.com/CQCL/guppylang) software libraries. While satisfying a lot of our intended criteria, HUGR itself was
```
ss2165, line 127:
```suggestion
[TKET](https://github.com/CQCL/tket2) and [Guppy](https://github.com/CQCL/guppylang) software libraries. While satisfying a lot of our intended criteria, HUGR itself was
```

we have naming problems...

docs/motivation.md
```diff
@@ -0,0 +1,91 @@
...
+
+*Jeff* was designed to solve a few very specific problems the field is currently encountering.
+
+## Better interdisciplinary support.
+
+**We need to be able to compile large, complex, quantum algorithms, and this is an interdisciplinary
+effort.**
+
+Compilation is an incredibly important component of quantum computing infrastructure. The inability
+to optimize large algorithms for future fault-tolerant hardware will extend(or block!) timelines
```
ss2165, line 11:
```suggestion
to optimize large algorithms for future fault-tolerant hardware will extend (or block!) timelines
```

docs/motivation.md
```diff
@@ -0,0 +1,91 @@
...
+Furthermore, as some compiler frameworks become more complex, making use of large classical
+toolchains such as LLVM and MLIR, the [overhead to contributing novel compilation passes is growing
+significantly](https://arxiv.org/abs/2411.18682). **We need to reduce this overhead while still
+allowing use of such tooling**.
+
+## Better visibility into software compilation.
+
+**We need to be able to visualize and debug our programs as they are modified by software compilers**.
+
+As compilers become more complex (using tooling such as Rust, or LLVM, as well as producing
```
ss2165, line 71: rust seems a touch out of place here

docs/motivation.md
```diff
@@ -0,0 +1,91 @@
...
+
+## There is still so much uncertainty.
+
+**We need to move fast, without slowing down development of new ideas.**
+
+The danger of introducing new exchange formats or standards is that they themselves may slow down
+development — perhaps because of a governance structure with too much process and/or stakeholders,
+or simply because the format is too rigid and doesn't permit extensibility for new ideas.
+
+This is something we wish to avoid — Jeff is *not* a standard, but an exchange format, designed to
```
ss2165, line 88: hmm you can have exchange format standards (e.g. JSON), maybe we want to say its not a standard we need all stakeholders to adopt since it is localised to just compilers choosing to reuse work?

ss2165 (commented):
the motivation section  is very cool 👍

docs/faqs.md
```diff
@@ -0,0 +1,131 @@
+# Frequently asked questions
+
+## Is *Jeff* meant to be an intermediate format that quantum software packages to manipulate?
```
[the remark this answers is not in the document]
josh146: potentially an unpopular opinion, but I'm not sure it matters too much 😅 The same way you see json and JSON interchangeably. But I saw that others have been using *Jeff*, so have been aiming to be consistent there

docs/faqs.md
```diff
@@ -0,0 +1,131 @@
...
+Moreover, OpenQASM3 has a few quirks that limit its general usability as a compiler exchange
+format:
+
+- Lack of extensibility without evolving the specification (a `#pragma` directive is provided, but
+  does not easily allow for including general structured information).
+
+- You can only index into arrays using compile-time constants, not dynamic values that may only been
+  known at runtime.
+
+- Registers (TODO).
```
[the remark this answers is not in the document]
josh146: I have it in my notes for QASM3, but I'm not sure either. I'll remove it for now.

docs/motivation.md
josh146:
```suggestion
This is something we wish to avoid — Jeff is *not* a standard we need all stakeholders to adopt, but an exchange format localized to compilers, designed to
```

docs/faqs.md
```diff
@@ -0,0 +1,144 @@
...
+  a lot of questions we need to consider and resolve when it comes to compiling dynamic, structured
+  programs. For example, the more compact the program representation (which is great for
+  compilation scalability!) the less information we potentially have up-front to maximally optimize
+  programs at compile time. How do we deal with such scenarios?
+
+## How did you decide what goes into the schema?
+
+## Why Cap'n Proto?
+
+Jeff is a machine-first exchange format, so we looked at existing succh projects.
```
mlxd, line 64:
```suggestion
Jeff is a machine-first exchange format.
```

Either this, or just fix the "succh->such"

docs/faqs.md
```diff
@@ -0,0 +1,145 @@
...
```
mlxd, line 64:
```suggestion
While we could have gone with something like JSON or MessagePack - we wanted first-class
```

docs/faqs.md
```diff
@@ -0,0 +1,145 @@
...
+While we could have gone with something like JSON or Messagepack - we wanted first-class
+schema support, efficient binary encoding and tooling in many languages.
+
+The [Protocol buffer](https://en.wikipedia.org/wiki/Protocol_Buffers) family of projects
+fits this description well. This includes the original 
+[Protocol buffer](https://protobuf.dev/) project but also more modern evolutions
```
mlxd, line 70: There's a lot more use of "protocol buffers" here than needed. Why not just mention "data serialization formats", or something generic, before introducing the current implementation using capnproto.

I'd reframe this as:
* We are using Capnproto
* Reasons we are using capnproto
* Alternative models to capnproto (the other models can be mentioned here)

docs/motivation.md
```diff
@@ -0,0 +1,91 @@
...
+*Jeff* was designed to solve a few very specific problems the field is currently encountering.
+
+## Better interdisciplinary support.
+
+**We need to be able to compile large, complex, quantum algorithms, and this is an interdisciplinary
+effort.**
+
+Compilation is an incredibly important component of quantum computing infrastructure. The inability
+to optimize large algorithms for future fault-tolerant hardware will extend (or block!) timelines
+for utility-scale quantum computation.
```
mlxd, line 12: Do we want to use this term?

docs/motivation.md
```diff
@@ -0,0 +1,91 @@
...
+- continue to make new theoretical compilation breakthroughs; 
+- identify the most-promising large scale algorithms; and finally,
+- figure out how to make this practical through quantum software.
+  
+To do so, we need to bridge the gap between researchers doing fundamental compilation and
+algorithmic research, and quantum software developers.
+
+- **How can we share results and ideas faster, within a common framework?**
+- **How do we make sure we are speaking a common language, and aware of challenges and approaches
+    on both sides?**
```
mlxd, line 25: These feel like they stand out here for no reason. Would it be worth adding lead-in or lead-out from the questions?

docs/motivation.md
```diff
@@ -0,0 +1,91 @@
...
+**To solve the above problems, we need better interoperability and visibility between compilers**.
+
+There are a wealth of quantum software compilers currently available, but they are not
+interoperable, and it is often hard to inspect/debug to see how your program changes. Furthermore,
+when *implementing* your own compilation pass, you often need to choose which framework to support,
+making it harder to share ideas and knowledge.
+
+Note that some formats have become de-facto exchange formats for compilers (such as OpenQASM), but
+were not designed to be exchange formats for compilers, and as such suffer from quantum
+representations that are not ideal for high-level optimization, lucklustre tooling, and
```
mlxd, line 59:
```suggestion
representations that are not ideal for high-level optimization, lacklustre tooling, and
```

josh146: thanks, I did stare at this word for a while wondering if something wasn't quite right

docs/faqs.md
```diff
@@ -0,0 +1,131 @@
...
+No. *Jeff* is designed purely to be an exchange format. It is expected that consuming software
+frameworks map **Jeff** input to their own intermediate representations for internal manipulation.
+
+## Why design a non-human readable format?
+
+The main goal is to design an *efficient* exchange format that best represents the information
+needed for quantum optimizations by compilers, and the most efficient approach is a binary,
+non-human readable format.
+
+[Protocol buffers](https://en.wikipedia.org/wiki/Protocol_Buffers) (and derivatives) are a binary
```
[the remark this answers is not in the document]
josh146:
```suggestion
[Protocol buffers](https://en.wikipedia.org/wiki/Protocol_Buffers) (and derivatives such as Cap'n Proto) are a binary
```

docs/motivation.md
josh146, line 25:
```suggestion
To do so, we need to bridge the gap between researchers doing fundamental compilation and
algorithmic research, and quantum software developers, and address the following challenges.

- **How can we share results and ideas faster, within a common framework?**
- **How do we make sure we are speaking a common language, and aware of challenges and approaches
    on both sides?**
```

docs/faqs.md
```diff
@@ -0,0 +1,131 @@
...
+  compilation scalability!) the less information we potentially have up-front to maximally optimize
+  programs at compile time. How do we deal with such scenarios?
+
+## How did you decide what goes into the schema?
+
+## Why Cap'n Proto?
+
+## Isn't this just *another* format standard?
+
+## Why not format X?
```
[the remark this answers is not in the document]
josh146: yep!

docs/faqs.md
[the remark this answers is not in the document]
ss2165: All the marketing used _jeff_ - also don't care too much

README.md
```diff
@@ -43,8 +43,10 @@ In particular,
 - Thinking beyond simple circuits will allow us to do new things and progress the field.
 
 Compilation and optimization of such programs is crucial to support quantum hardware execution,
-however current de-facto interchange formats (such as QIR and OpenQASM) either aren't designed for quantum optimization — lacking the explicit
-structure needed for advanced high-level optimizations — nor tool-agnostic exchange, locking developers into particular toolchains like LLVM.
+however current de-facto interchange formats (such as QIR and OpenQASM) either aren't designed for
+quantum optimization — lacking the explicit structure needed for advanced high-level
+optimizations — nor tool-agnostic exchange, locking developers into particular toolchains like
+LLVM.
```
glassnotes, line 49: Suggest rephrasing since "quantum optimization" is ambiguous - I assume what is meant is quantum program optimization and not quantum algorithms for optimization. But also, I don't think it's the case that they aren't designed for compilation/optimization, it's more that the format is amenable only to optimization of the lowest-level (gate) instructions.

```suggestion
however current de-facto interchange formats (such as QIR and OpenQASM) either lack the explicit structure needed for advanced high-level
optimizations, or are not tool-agnostic, locking developers into particular toolchains like
LLVM.
```

README.md
```diff
@@ -43,8 +43,10 @@ In particular,
...
 
 ***This creates a bottleneck for collaboration and makes it difficult to develop and share
```
glassnotes, line 51: [I can't comment on the collapsed lines (54-56), so leaving a comment here.]  

It would be helpful to provide some more motivation as to why a serialized, in-memory representation is useful (presumably w.r.t. speed and scale?). 

EDIT: I see this is covered in the FAQs but putting it up front wouldn't hurt.

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Why design a non-human readable format?
+
+The main goal is to design an *efficient* exchange format that best represents the information
+needed for quantum optimizations by compilers, and the most efficient approach is a binary,
+non-human readable format.
+
+[Protocol buffers](https://en.wikipedia.org/wiki/Protocol_Buffers) (and derivatives such as Cap'n Proto) are a binary
+data format initially developed by Google that has been used for similar problems across the
+classical domain, and was chosen for *Jeff*.
```
glassnotes, line 16:
```suggestion
data format initially developed by Google that have been used for similar problems across the
classical domain, and were chosen for *Jeff*.
```

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+Human-readable formats have been quite common in the quantum space thus far, but in addition we see
+two major reason why they might make less sense going forward:
+
+- We are only just beginning to scale quantum software to FTQC regimes, so have not experienced the
+  performance issues associated with (re-)parsing human readable formats, compared to formats
+  designed for machine-readability.
+
+- the tooling around IR and exchange formats has not been mature enough, and the best way to
+  debug/inspect has often been to simply open and read the file. Improving tooling for debugging
+  and visualization should help rectify this.
```
glassnotes, line 27: Is the idea here that a debugging tool might also parse Jeff, and display some structured program representation?  Would that require either (a)  tools to be implemented by the individual compilers / providers, or (b) tools that allow visualization and/or debugging at a universal but more abstract level?

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Why do we need program structure?
+
+There are two main reasons:
+
+- Because we *need* to (not all algorithmic primitives, an example being repeat-until-success, can
```
glassnotes, line 33: Minor point but the grammar of the response doesn't match with the question in the header

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Why Cap'n Proto?
+
+We use [Cap'n Proto](https://capnproto.org/) as our data serialization format for Jeff.
+
+**Reasons for choosing Cap'n Proto:**
+- Fast, efficient binary encoding
+- Zero-copy deserialization for high performance
+- Simple and expressive schema language
+- Robust tooling and support for key quantum programming languages (Python, C++, Rust)
```
glassnotes, line 70:
```suggestion
- Robust tooling and support for key quantum programming frameworks (Python, C++, Rust)
```

I don't think the compilers/frameworks are formal languages, but tbh I don't know enough to explain precisely where the line is drawn.

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+
+The danger of introducing new exchange formats or standards is that they themselves may slow down
+development — perhaps because of a governance structure with too much process and/or stakeholders,
+or simply because the format is too rigid and doesn't permit extensibility for new ideas.
+
+This is something we wish to avoid — Jeff is *not* a standard we need all stakeholders to adopt, but
+an exchange format localized to compilers, designed to
+be easily extensible. Additional information for program representation and compilation can be added
+directly, and will be taken into account by software frameworks that understand the direction.
+Further, development will happen out in the open, on GitHub.
```
glassnotes, line 86: I'm not convinced that open development alone will avoid the problems mentioned here. Is there a broader strategy to get _feedback_ from stakeholders and/or competing IRs, without binding them to use the format or requesting them to steer development?

As a side note... I've facilitated discussions at compilation workshops to try and determine what we need / don't need as part of a unified representation. These discussions never converge :sweat_smile: There is always something missing that someone wants. An important question to consider is, does there exist a _truly minimal_ set of requirements for an exchange format such that any (known) quantum algorithm can be compiled to Jeff from framework A, and faithfully loaded into framework B?

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+compiler exchange format.
+
+An exchange format that makes it easy to incorporate structure will also allow the field to consider
+compilation of structured programs by default.
+
+### Why not OpenQASM3?
+
+[OpenQASM3](https://arxiv.org/abs/2104.14722) brings program structure to OpenQASM2, however has not yet seen adoption or tooling
+maturity in the ecosystem to the extent as OpenQASM2. In addition, OpenQASM3 maintains a focus on
+human readability (requiring string generation and parsing), and is not designed first and foremost for quantum optimization.
```
dime10, line 103: I think the two main disadvantages should be placed in the bulleted list for better visibility. To me these are the reliance on parsing, and the linear arrangement of gates rather than an arrangement in a graph.

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+[CUDA-Q](https://github.com/nvidia/cuda-quantum) and [Catalyst](https://github.com/pennylaneai/catalyst) — rather than compiling QIR programs, they instead
+optimize MLIR, a higher-level intermediate representation.
+
+Furthermore, relying on QIR as a compilation exchange format introduces a big barrier to entry due
+to it depending on the LLVM toolchain — a massive software project that has non-trivial
+compatibility concerns, installation, and learning overhead.
+
+### Why not MLIR?
+
+[MLIR (Multi-Level Intermediate Representation)](https://en.wikipedia.org/wiki/MLIR_(software)) is a framework that exists on top of QIR/LLVM, and
```
dime10, line 134: MLIR doesn't exist on top of QIR, you could argue it exists on top of LLVM (a lof of inspiration is taken from llvm, mlir uses a lot of core support libraries from llvm, and there is built-in lowering from standard mlir dialects to llvm ir), although even that is not necessarily a conceptual requirement for mlir (it can be used without going to llvm ir, or implemented without llvm libraries like in xdsl).

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+permits multiple high-level 'dialects' (IRs tailored to specific domains) to coexist, making it the
+framework of choice for most QIR/LLVM compiler stacks such as CUDA-Q and Catalyst.
```
dime10, line 136: I would not call Catalyst a QIR compiler stack, it doesn't have support for emitting something reasonably close to the QIR spec.

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+
+However, something to note is that MLIR is a *framework*, not a defined format — different compilers
+define *multiple* internal/custom dialects that altogether are used to represent a quantum program.
+At the moment, there is no clear *shared* set of quantum MLIR dialects that make sense to use more
```
dime10, line 140: An additional point to mention is the difficulty in maintaining binary compatibility between different projects, even if they did share the same mlir dialect. The textual mlir format would avoid this though.

README.md
josh146: Good point, might make a slight modification here:

```suggestion
however current de-facto interchange formats (such as QIR and OpenQASM) either lack the explicit structure needed for advanced high-level
quantum optimizations, or are not tool-agnostic, locking developers into particular toolchains like
LLVM.
```

docs/faqs.md
josh146: https://github.com/unitaryfoundation/jeff/pull/10/files#r2312562198

docs/faqs.md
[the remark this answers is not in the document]
josh146:
```suggestion
```

docs/faqs.md
josh146:
```suggestion
[MLIR (Multi-Level Intermediate Representation)](https://en.wikipedia.org/wiki/MLIR_(software)) is a framework that developed by the LLVM project, and
```

docs/faqs.md
josh146:
```suggestion
framework of choice for most LLVM compiler stacks such as CUDA-Q and Catalyst.
```

docs/faqs.md
```diff
@@ -0,0 +1,153 @@
...
+broadly.
```
josh146, line 141:
```suggestion
broadly.

Further, even if projects do share the same MLIR dialect, an
additional difficulty would be maintaining binary compatibility
between different projects.
```

docs/faqs.md
josh146: https://github.com/unitaryfoundation/jeff/pull/10/files#r2312563136

README.md
josh146: I might address this in a subsequent PR!

docs/faqs.md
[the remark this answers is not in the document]
dime10: > All the marketing used _jeff_ - also don't care too much

That's what I've been using in any formatted text since that was the only vote received at the time^^ I'm still not sure what I like better though

aborgna-q (approved):

docs/faqs.md
glassnotes: Ah it was this flow: "Why do we need program structure?" -> "Because we need to"

PR #11, opened 2025-08-28
Add image of structured quantum programs

josh146:


ss2165 (commented):
not sure what the intended behaviour is but on dark mode I see this (no text) 

Image

and on light mode nothing loads


perhaps we should use the alt text also?

josh146:
>> not sure what the intended behaviour is but on dark mode I see this (no text)
>
> Is the intent to hand-wavingly show increasing program structure going from left to right? If so, I think its a good image, but perhaps a caption or quick comment (and alt text) to relate that would help?

Yep, exactly. There's no text in the image, but it is an abstract representation.

I wanted to avoid implying circuits = bad, since circuits _are_ useful and do have their place. Moreso I wanted to emphasize that circuits are just one of many representations we need to consider, alongside more general structured representations.

Good call on adding a caption 👍

PR #12, opened 2025-08-29
ci: Add separate name for 🦀/📜 required checks steps

aborgna-q:
It looks like github required checks only look for the _step name_, without discriminating between workflow names.

Since we have "Required checks" steps on each of the python, rust, and schema workflows we need to ensure they all have different names. (Python already had a 🐍 suffix).


bachase (approved):
LGTM -- I can make the corresponding change in protection rules once this is merged.

PR #23, opened 2025-09-04
chore: Run pytest and ruff check on example notebooks

aborgna-q:


burgholzer (approved):
It's been a while since this PR was last touched, but this LGTM.
@aborgna-q does it still make sense to merge this?

PR #25, opened 2025-09-04
test: Check python notebooks with pytest

aborgna-q:


burgholzer (approved):
Also feels like this would be ready to merge to me.

PR #32, opened 2025-11-02
rfc: collect and provide structured benchmark programs

DRovara:
This PR proposes RFC 0032, which advocates for the addition of a benchmark suite for structured quantum programs.

## Problem

There is no standard benchmark suite for evaluating compiler support for structured control flow (e.g., if, for, dynamic indexing).

## Solution

Create a set of benchmarks written in Jeff to fill this gap. This will help drive compiler development, allow for better tool evaluation, and highlight Jeff's capabilities in representing these advanced programs.

This RFC outlines the initial set of benchmarks and invites community feedback.

[Rendered RFC](https://github.com/DRovara/jeff/blob/rfc/structured-benchmark-programs/rfcs/text/0032-structured-benchmark-programs.md)

### DIFF
--- a/rfcs/text/0032-structured-benchmark-programs.md
+++ b/rfcs/text/0032-structured-benchmark-programs.md
@@ -201,7 +201,7 @@ There are several potential drawbacks to consider with this proposal:
 - Maintenance Overhead: The addition of a new benchmark suite requires ongoing maintenance to ensure that the benchmarks remain relevant and up-to-date with the latest advancements in quantum computing and compiler technologies.
 - Complexity: Introducing structured benchmarks may increase the complexity of the Jeff repository, potentially making it more challenging for new users to navigate and understand the available resources.
 - Limited Adoption: If the benchmarks are not widely adopted by the quantum computing community, their impact may be limited, reducing the incentive for compiler developers to implement support for structured control flow.
-- Comparisons are not simple: For users, it might not be stratighforward to know what to compare against when compiling these structured programs. For a full, fair evaluation, a meaningful baseline needs to be established and a more precise methodology for comparison is likely necessary.
+- Comparisons are not simple: For users, it might not be straightforward to know what to compare against when compiling these structured programs. For a full, fair evaluation, a meaningful baseline needs to be established and a more precise methodology for comparison is likely necessary.
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives


rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,157 @@
...
+
+*Note: this list was created from previous discussions among community members and is open for further suggestions and modifications.*
+
+## Static Loops without Dynamic Qubit Indexing
+
+*This class of benchmark programs utilizes loops (such as `for` and `while`) that **do not** depend on dynamic execution results. Inside the loop bodies, all operations are still statically indexed.*
+
+- *Grover's Search Algorithm*: Oracle and diffusion operations can be applied repeatedly using a fixed-size loop. (https://arxiv.org/abs/quant-ph/9605043)
+- *VQE Ansatz with Fixed Repetitions*: A variational ansatz circuit that applies a set of parameterized gates in a loop with a predetermined number of repetitions. ([Peruzzo et al., 2014](https://arxiv.org/abs/1304.3061))
+- *QAOA with Fixed Repetitions*: A Quantum Approximate Optimization Algorithm circuit that applies problem and mixer Hamiltonians in a loop with a fixed number of layers. ([Farhi et al., 2014](https://arxiv.org/abs/1411.4028))
```
burgholzer: In both of these cases, the loop body would not index any qubits, but it would index into some parameter array, right? I suppose that is supported by pretty much every framework/language, right?
Just want to make sure that these are really in the same category as the Grover example.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,157 @@
...
+# Implementation-level explanation
+[implementation-level-explanation]: #implementation-level-explanation
+
+The following list outlines the proposed structured benchmark programs to be included in the `benchmarks/structured` directory, along with brief description, grouped by the way they utilize structured control flow:
+
+*Note: this list was created from previous discussions among community members and is open for further suggestions and modifications.*
+
+## Static Loops without Dynamic Qubit Indexing
+
+*This class of benchmark programs utilizes loops (such as `for` and `while`) that **do not** depend on dynamic execution results. Inside the loop bodies, all operations are still statically indexed.*
```
burgholzer: Is the term "dynamic execution results" established enough in quantum?
I would have probably went into the direction of "do not use mid-circuit measurements and conditional quantum operations", or similar.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,157 @@
...
+
+- *Grover's Search Algorithm*: Oracle and diffusion operations can be applied repeatedly using a fixed-size loop. (https://arxiv.org/abs/quant-ph/9605043)
+- *VQE Ansatz with Fixed Repetitions*: A variational ansatz circuit that applies a set of parameterized gates in a loop with a predetermined number of repetitions. ([Peruzzo et al., 2014](https://arxiv.org/abs/1304.3061))
+- *QAOA with Fixed Repetitions*: A Quantum Approximate Optimization Algorithm circuit that applies problem and mixer Hamiltonians in a loop with a fixed number of layers. ([Farhi et al., 2014](https://arxiv.org/abs/1411.4028))
+
+## Static Loops with Dynamic Qubit Indexing
+
+*This class of benchmark programs utilizes loops (such as `for` and `while`) that **do not** depend on dynamic execution results. Inside the loop bodies, operations may be dynamically indexed (e.g. based on the loop parameter).*
+
+- *GHZ State Preparation*: A static loop is used to entangle all qubits with each other using `cx` gates where, at each loop index `i`, the `cx` gate is applied between qubit `0` and qubit `i`. ([Wikipedia on GHZ states](https://en.wikipedia.org/wiki/GHZ_state))
```
burgholzer: Just noting this here: An alternative construction for the GHZ state would be to entangle `i` with `i+1` for `i` from 0 to n-2.
Gives a linear topology circuit compared to a star topology circuit.
Might be worth to include both?

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,157 @@
...
+- *Quantum Fourier Transform (QFT)*: Nested static loops are used to apply controlled phase rotations between qubits, where the control and target qubits are determined based on the loop index. ([Nielsen and Chuang, 2010](https://doi.org/10.1017/CBO9780511976667))
+- *Quantum Phase Estimation (QPE)*: Static loops are used to apply controlled unitary operations and inverse QFT, with qubit indices determined by the loop parameters. ([Nielsen and Chuang, 2010](https://doi.org/10.1017/CBO9780511976667))
```
burgholzer: Although it's just a case of QPE, we could think about including Shor's algorithm in this list here.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,157 @@
...
+
+- *Quantum Multiplexers*: Quantum multiplexers apply different unitary operations based on the state of control qubits, effectively implementing conditionals in quantum circuits. ([Quantum multiplexers in circuit synthesis](https://arxiv.org/abs/quant-ph/0410066))
+
+## Dynamic Qubit (Re-)Allocation
+
+*This class of benchmark programs allocates qubits dynamically during execution, based on runtime conditions or within loop bodies.*
+
+- *Iterative Quantum Fourier Transform (iQFT)*: Rather than allocating all qubits upfront, just a single qubit is allocated at the start of execution and then reset during each loop iteration of QFT.
+- *Iterative Quantum Phase Estimation (iQPE)*: Similar to iQFT, iQPE allocates a single qubit (in addition to the phase qubit "psi") and reuses it across iterations to estimate the phase of a unitary operator.
+- *Shor's Algorithm*: The primitives used in Shor's algorithm can similarly be implemented using dynamic qubit reallocation to minimize the total number of qubits required.
```
burgholzer: I think there is another class of algorithms that could be really interesting here that truly dynamically allocates qubits and deallocates them again and that is the category of programs heavily using arithmetic circuit components. Most of these components (think adders or multipliers) heavily use multi-controlled Toffoli gates in their descriptions. These gates benefit substantially from the availability of ancillary qubits for their decomposition to native gates. Such ancillary qubits may be allocated on demand and depending on availability.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Loosely Coupled Hybrid Programs
+
+*Variational and hybrid algorithms naturally combine quantum circuits with classical optimization loops. They are loosely coupled because the quantum and classical parts interact only at well-defined synchronization points.*
+
+- *VQE*: A variational quantum eigensolver algorithm that uses a classical optimizer to minimize the energy of a parameterized quantum circuit which implements several iterations of some ansatz. ([Peruzzo et al., 2014](https://arxiv.org/abs/1304.3061))
+- *QAOA*: A quantum approximate optimization algorithm that uses a classical optimizer to find optimal parameters for a parameterized quantum circuit. It iteratively applies a "problem" and "mixer" operator using loop structures. ([Farhi et al., 2014](https://arxiv.org/abs/1411.4028))
```
burgholzer: We might want to explicitly declare this as out-of-scope for this RFC.
The main difference here is that the quantum-classical interaction does _not_ have to happen within the coherence time of the qubits.
This might be an important distinction to make more broadly when we refer to structured programs onwards.

burgholzer (changes requested):
Hey @DRovara 👋🏼 

Thanks for kicking this off! Great to see the first RFC ;-)
I went through the document top-to-bottom once and accumulated some comments. Most of them are pretty minor and should be fairly easy to address.

I have one more general comment or request for changes:
Would it make sense to define a list of "defining features" for a structured program and then create a table for the benchmarks that highlights which features a particular program uses?
Something like:
- Loops with compile-time bounds
- Loops with runtime bounds (true `while` loops)
- Dynamic qubit indexing (within loops)
- Conditional quantum instructions (e.g., depending on mid-circuit measurement results)
- Qubit reuse (e.g., via `reset` instructions)
- Dynamic qubit allocation

This might make it a little easier to handle the fact that it's hard to put some of the algorithms into a single category.
The table could also include a column with a short description and a column for references to each algorithm as well as, potentially, a checkbox that could be ticked once the algorithm is implemented.
This could create a more structured (pun intended) way of organizing the different benchmarks than the current section-based layout.

rfcs/text/0032-structured-benchmark-programs.md
DRovara: That's a good point that I haven't really thought of. Considering that OpenQASM 3 also supports loops like these:
```
array[float[64], 4] my_floats = {1.2, -3.4, 0.5, 9.8};
for float[64] f in my_floats {
   // do something with 'f'
}
```

I assume the VQA loops might be expressed in that way, as in iterating over an array of float arrays where each individual array contains the angles for a single repetition layer.

rfcs/text/0032-structured-benchmark-programs.md
DRovara: Added that algorithm class to the list

rfcs/text/0032-structured-benchmark-programs.md
burgholzer: Would be interesting to know if SDKs like Qiskit can handle such loops when importing OpenQASM 3 programs.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,162 @@
+- Feature Name: structured-benchmark-programs
+- Start Date: 2025-10-31
+- RFC PR: [unitaryfoundation/jeff#0032](https://github.com/unitaryfoundation/jeff/pull/0032)
+
+# Summary
+[summary]: #summary
+
+Quantum programs involving structured control flow operations such as `if` and `for` can only be represented in severely limited ways through existing quantum program formats. Qiskit does not allow dynamic qubit indexing, which is crucial for semantic `for` loops. Similarly, while OpenQASM 3 does not explicitly prohibit dynamic indexing, it leaves it open as an optional feature that many backends do not support. Jeff, on the other hand, has the ability to represent such programs, making it the ideal format for expressing a set of structured benchmark programs. These programs can then be used as a set of challenges for existing quantum compilers to drive the development of more advanced compilation techniques.
```
josh146: nice!

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,192 @@
...
+
+This chapter outlines the proposed structured benchmark programs to be included in the `benchmarks/structured` directory.
+In the table below, for each program type, the required structured control flow features are listed, as well as references to relevant literature or resources for further reading.
+Programs for which "*arbitrary size*" is marked can be defined generically and parameterized by input parameters to allow for different sizes at runtime.
+Further details on each program type are provided in the sections that follow the table.
+
+[!NOTE]
+This list was created from previous discussions among community members and is open for further suggestions and modifications.*
+
+| Program Type | statically-bounded loops | dynamically-bounded loops | dynamic qubit indexing | conditionals | dynamic qubit allocation | qubit reuse | references | arbitrary-size |
```
glassnotes: This is a great table :100:

rfcs/text/0032-structured-benchmark-programs.md
glassnotes: I was confused by this as well. I'm leaving a related comment in the "Conditionals" section.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Dynamic Qubit (Re-)Allocation
+
+*This class of benchmark programs allocates qubits dynamically during execution, based on runtime conditions or within loop bodies.*
+
+- *Iterative Quantum Fourier Transform (iQFT)*: Rather than allocating all qubits upfront, just a single qubit is allocated at the start of execution and then reset during each loop iteration of QFT.
+- *Iterative Quantum Phase Estimation (iQPE)*: Similar to iQFT, iQPE allocates a single qubit (in addition to the phase qubit "psi") and reuses it across iterations to estimate the phase of a unitary operator.
+- *Shor's Algorithm*: The primitives used in Shor's algorithm can similarly be implemented using dynamic qubit reallocation to minimize the total number of qubits required.
+- *Toffoli-heavy Circuits*: Toffoli gates in quantum circuits can be implemented more efficiently using ancilla qubits. Circuits using a larger number of Toffoli gates, such as many arithmetic circuits, can benefit from dynamic qubit allocation to allocate and deallocate ancilla qubits repeatedly inside loop bodies.
```
glassnotes: Another example could be the parallelization technique applied in [Quantum Fan-out is Powerful](https://www.theoryofcomputing.org/articles/v001a005/v001a005.pdf). See Figs. 3 and 4.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,192 @@
...
+## Loosely Coupled Hybrid Programs
+
+*Variational and hybrid algorithms naturally combine quantum circuits with classical optimization loops. They are loosely coupled because the quantum and classical parts interact only at well-defined synchronization points.*
+
+- *VQE*: A variational quantum eigensolver algorithm that uses a classical optimizer to minimize the energy of a parameterized quantum circuit which implements several iterations of some ansatz. ([Peruzzo et al., 2014](https://arxiv.org/abs/1304.3061))
+- *QAOA*: A quantum approximate optimization algorithm that uses a classical optimizer to find optimal parameters for a parameterized quantum circuit. It iteratively applies a "problem" and "mixer" operator using loop structures. ([Farhi et al., 2014](https://arxiv.org/abs/1411.4028))
+
+## Fault-Tolerant State Preparation and Error Correction
+
+*Quantum Error Correction (QEC) is one of the most important applications of structured control flow in quantum computing. These benchmark programs implement QEC protocols that involve structured operations at any point in the program.*
```
glassnotes: As I've gone down the list, I wonder if it's worth making a distinction between "simple" benchmark that utilize a single structured primitive, and "compound" benchmarks that combine multiple primitives.

For example, QFT is a simple benchmark that uses a dynamically-sized loop that indexes qubits. QPE is compound in the sense that it does this in two places (controlled rotations, plus the QFT). Something like error correction is really complex as it weaves many primitives together.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,162 @@
...
+# Summary
+[summary]: #summary
+
+Quantum programs involving structured control flow operations such as `if` and `for` can only be represented in severely limited ways through existing quantum program formats. Qiskit does not allow dynamic qubit indexing, which is crucial for semantic `for` loops. Similarly, while OpenQASM 3 does not explicitly prohibit dynamic indexing, it leaves it open as an optional feature that many backends do not support. Jeff, on the other hand, has the ability to represent such programs, making it the ideal format for expressing a set of structured benchmark programs. These programs can then be used as a set of challenges for existing quantum compilers to drive the development of more advanced compilation techniques.
+
+# Motivation
+[motivation]: #motivation
+
+Many programs can be greatly simplified through the use of structured control flow operations.
+This includes simple loops for repeated operations, conditionals to implement classically controlled quantum operations in a natural way, and more complex loops and control flow structures that depend on measurement results and access qubits dynamically.
```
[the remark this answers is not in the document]
DRovara: That's a very good point, I have now added a paragraph on that.

rfcs/text/0032-structured-benchmark-programs.md
DRovara: Just did some checking:
Qiskit does not support arrays at all so I guess there is no way to do this in qiskit after all.
I also tried it with Q# since parsing OpenQASM code is pretty easy there and Q# actually managed to parse a program like that.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,196 @@
...
+However, a large number of existing quantum compiler toolchains still do not support some or even all of these features.
+Proposing a set of such programs as benchmarks can not only help users evaluate the capabilities of different toolchains, but also serve as a motivation for compiler developers to implement features such as the compilation of such programs *without unrolling* the program structure, and even *rethink the use of circuits* as the default program representation for quantum compilation - can they do better and result in more optimized programs by taking into account the additional information available from the structure?
+
+Furthermore, it can also increase the visibility of Jeff as a quantum program format that can express these advanced features, encouraging more users to adopt it for their quantum programming needs.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A new directory, `benchmarks/structured` is introduced in the Jeff repository.
+It provides a set of individual Jeff files that each represent a single structured benchmark program.
```
mark-koch: Similar to what @burgholzer brought up [here](https://github.com/unitaryfoundation/jeff/pull/32/files#r2485571399), I think the repo should also contain the source code that was used to produce the jeff binaries. This might differ from example to example (e.g. Catalyst, Guppy, MQT, Jeff python builder, Hugr builder, etc), but if we don't include the source we'll have a hard time regenerating the binaries e.g. when updating the format

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,196 @@
...
+## Dynamic Loops
+
+*This class of benchmark programs utilizes loops (such as `for` and `while`) that **do** depend on dynamic execution results, such as measurement outcomes. Operations may also be dynamically indexed.*
+
+- *Grover's Search with Weak Measurement*: Measurements on an ancilla qubit are utilized to determine whether to continue with another iteration of Grover's search or to stop as amplitude amplification has succeeded. ([Andrés-Martínez et al.](https://iopscience.iop.org/article/10.1088/2058-9565/ac47f1/meta))
+- *Repeat-Until-Success*: Specific operations are applied repeatedly until a desired measurement outcome is achieved, indicating that a desired complex operation has been successfully implemented.
+- *Quantum Metropolis Sampling*: [Temme et al., 2011](https://arxiv.org/abs/0911.3635)
+- *ML-QAE*: [Suzuki et al., 2020](https://arxiv.org/abs/1904.10246)
+
+## Conditionals
```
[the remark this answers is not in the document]
josh146: nice idea!

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There are several potential drawbacks to consider with this proposal:
+
+- Maintenance Overhead: The addition of a new benchmark suite requires ongoing maintenance to ensure that the benchmarks remain relevant and up-to-date with the latest advancements in quantum computing and compiler technologies.
+- Complexity: Introducing structured benchmarks may increase the complexity of the Jeff repository, potentially making it more challenging for new users to navigate and understand the available resources.
+- Limited Adoption: If the benchmarks are not widely adopted by the quantum computing community, their impact may be limited, reducing the incentive for compiler developers to implement support for structured control flow.
+- Comparisons are not simple: For users, it might not be stratighforward to know what to compare against when compiling these structured programs. For a full, fair evaluation, a meaningful baseline needs to be established and a more precise methodology for comparison is likely necessary.
```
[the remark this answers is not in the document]
glassnotes: Yeah, I think quantitative metrics + evaluation deserves a separate discussion.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,192 @@
...
+## Dynamic Loops
+
+*This class of benchmark programs utilizes loops (such as `for` and `while`) that **do** depend on dynamic execution results, such as measurement outcomes. Operations may also be dynamically indexed.*
+
+- *Grover's Search with Weak Measurement*: Measurements on an ancilla qubit are utilized to determine whether to continue with another iteration of Grover's search or to stop as amplitude amplification has succeeded. ([Andrés-Martínez et al.](https://iopscience.iop.org/article/10.1088/2058-9565/ac47f1/meta))
+- *Repeat-Until-Success*: Specific operations are applied repeatedly until a desired measurement outcome is achieved, indicating that a desired complex operation has been successfully implemented.
+- *Quantum Metropolis Sampling*: [Temme et al., 2011](https://arxiv.org/abs/0911.3635)
+- *ML-QAE*: [Suzuki et al., 2020](https://arxiv.org/abs/1904.10246)
+
+## Conditionals
```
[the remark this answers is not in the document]
glassnotes: >  In what degree should "quantum conditionals" really be a thing?

Right... the context for the comment is some early work on [quantum random access machines](https://link.springer.com/chapter/10.1007/978-3-031-02516-7_5#preview) (see Ch. 5.2) and Knill's [Conventions for Quantum Pseudocode](https://arxiv.org/abs/2211.02559). These works distinguish between classical and quantum conditionals (sometimes you'll even see `qif`). `qif` can serve as pseudocode for controlled operations, whereas controlling on a measurement outcome is like an in-between case. In that sense, I  view a multiplexer as like a quantum switch statement (where the switch applies in superposition).

In any case, these are unnecessary details, the distinction in the table is sufficient 😄 (I just think it's an interesting topic!)

rfcs/text/0032-structured-benchmark-programs.md
bachase: I like this idea as well (as also listed in other comments). For the purposes of this RFC, would it help to layout a bit more of that contribution process? 

Even if we wait a bit and get a bottoms up approach, it's unclear to me  what that initial contribution might look like. For say a parameterized circuit, would it be Jeff python builder code? Would we store the generated output for some specific parameter values, or only store the Jeff builder code in the repo?

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,224 @@
...
+- [UCC Bench](https://github.com/unitaryfoundation/ucc-bench): A command-line utility designed to benchmark and compare the performance of various quantum compilers, with a particular focus on the ucc compiler.
+- [MQT Bench](https://github.com/munich-quantum-toolkit/bench): A collection of benchmark circuits in qiskit format to evaluate quantum software tools.
+- [metriq-gym](https://github.com/unitaryfoundation/metriq-gym): A framework for implementing and running standard quantum benchmarks.
+
+While many of these benchmark suites are widely used and provide valuable insights into quantum compiler performance, they often lack support for structured control flow features such as dynamic loops and conditionals.
+
+To the best of our knowledge, there is currently no widely adopted benchmark suite that specifically targets the evaluation of quantum compilers' support for structured control flow operations, although the topic has been discussed in various community gatherings and conferences.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
```
bachase: Process question -- are these blocking and need to be resolved prior to accepting the RFC and starting work?

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,224 @@
...
+---
+
+Each of the benchmark programs listed above should be implemented individually to achieve a comprehensive suite that covers a wide range of structured control flow scenarios.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There are several potential drawbacks to consider with this proposal:
+
+- Maintenance Overhead: The addition of a new benchmark suite requires ongoing maintenance to ensure that the benchmarks remain relevant and up-to-date with the latest advancements in quantum computing and compiler technologies.
```
bachase: We can also spin up a separate jeff-bench repo or similar, at least to avoid any challenges managing dependencies for the core jeff code from the benchmark generation code.

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,224 @@
...
+### Category Details
+
+| Category | Description |
+|----------|-------------|
+| statically-bounded loops                      | Program uses loops with a constant number of repetitions. |
+| dynamically-bounded loops                     | Loop bounds depend on value calculated at runtime. |
+| dynamic qubit indexing                        | Gates are applied to qubits with non-constant indices (e.g. loop variable). |
+| dynamic classical values                      | Gates use other classical values computed at runtime (e.g. rotation angles taken from arrays). |
+| conditionals on originally classical values   | Conditional blocks are used where the condition depends on values that were *not* measurement results. |
+| conditionals on measurement results           | Conditional blocks are used where the condition depends on values that depend on measurement results. |
```
dime10: I'm curious for the conditional we separate out classical dynamic values from "quantum dynamic values" (i.e. measurement results), but we didn't do this for the loops (I think both are contained in the dynamically-bounded loops).

rfcs/text/0032-structured-benchmark-programs.md
dime10: Dynamic program values instead of dynamic execution results might be a bit clearer imo, or something runtime-computed program values.

dime10 (commented):
Thanks Damian, this is a great first RFC!

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,224 @@
...
+|-----|--------------|-------------------------|--------------------------|-----------------------|--------------|-------------------------|-------------|-----------|--------|----------|----------|
+| Quantum Teleportation                         | ❌ | ❌ | ❌ | ❌ | ❌ | ✔️ | ❌ | ❌ | [Paper](https://doi.org/10.1103/PhysRevLett.70.1895) | ❌ | ❌ |
+| Block Encoding                                | ❌ | ❌ | ❌ | ❌ | ❌ | ✔️ | ❌ | ❌ | [Paper](https://arxiv.org/abs/1606.02685), [Paper](https://arxiv.org/abs/1806.01838) | ✔️ | ❌ |
+| Grover's Search Algorithm                     | ✔️ | ❌ | ❌ | ❌ | 🟦 | ❌ | ❌ | ❌ | [Paper](https://arxiv.org/abs/quant-ph/9605043) | ✔️ | ❌ |
+| Grover's Search with Weak Measurement         | ❌ | ✔️ | ❌ | ❌ | 🟦 | ✔️ | ❌ | ❌ | [Paper](https://iopscience.iop.org/article/10.1088/2058-9565/ac47f1/meta) | ✔️ | ✔️ |
+| GHZ State Preparation                         | ✔️ | ❌ | ✔️ | ❌ | ❌ | ❌ | ❌ | ❌ | [Wikipedia](https://en.wikipedia.org/wiki/GHZ_state) | ✔️ | ❌ |
+| Quantum Fourier Transform (QFT)               | ✔️ | ❌ | ✔️ | 🟦 | ❌ | ❌ | ❌ | ❌ | [Nielsen and Chuang](https://doi.org/10.1017/CBO9780511976667) | ✔️ | ❌ |
```
glassnotes:
```suggestion
| Quantum Fourier Transform (QFT)               | ✔️ | ❌ | ✔️ | 🟦 | ❌ | ❌ | ❌ | ❌ | [Nielsen and Chuang](https://doi.org/10.1017/CBO9780511976667) | ✔️ | ❌ |
| QFT adder (quantum input, two registers)     | ✔️ | ❌ | ✔️ | 🟦 | ❌ | ❌ | ❌ | ❌ | [Draper](https://arxiv.org/abs/quant-ph/0008033) | ✔️ | ✔️ |
| QFT adder (classical input, single register)    | ✔️ | ❌ | ✔️ | ✔️|  🟦 | ❌ | ❌ | ❌ | [Beauregard](https://arxiv.org/abs/quant-ph/0205095), Fig. 3 | ✔️ | ✔️ |
| Controlled multiplication modulo N    | ✔️ | ✔️ | ✔️ | ✔️|  🟦 | ❌ | ❌ | ❌ | [Beauregard](https://arxiv.org/abs/quant-ph/0205095), Fig. 6 | ✔️ | ✔️ |
```

Would appreciate another pair of eyes on the criteria. 

The main thing is for the QFT adder with classical input (and controlled multiplication, which leverages it), loop bounds and rotation amounts are derived from input integers N (the modulus) and b (the number to be added), whereas the regular two-register QFT adder is a sequence of controlled rotations from one register to another.

rfcs/text/0032-structured-benchmark-programs.md
glassnotes: Also added to table; can resolve this and decide above whether it's worth including.

rfcs/text/0032-structured-benchmark-programs.md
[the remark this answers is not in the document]
mark-koch: I think the table entry for MBQC should look like

```
| Measurement-based quantum computation | 🟦 | ❌ | 🟦 | ✔️ | ❌ | ✔️ | ❌ | 🟦 | [Wikipedia](https://en.wikipedia.org/wiki/One-way_quantum_computer) | ❌ | 🟦 |
```

The main defining feature is measurement feed-forward and potentially dynamic-angle basis measurements. If we want to, we could also roll examples into statically bounded loops with qubit indexing, but not strictly necessary imo

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,224 @@
...
+| Category | Description |
+|----------|-------------|
+| statically-bounded loops                      | Program uses loops with a constant number of repetitions. |
+| dynamically-bounded loops                     | Loop bounds depend on value calculated at runtime. |
+| dynamic qubit indexing                        | Gates are applied to qubits with non-constant indices (e.g. loop variable). |
+| dynamic classical values                      | Gates use other classical values computed at runtime (e.g. rotation angles taken from arrays). |
+| conditionals on originally classical values   | Conditional blocks are used where the condition depends on values that were *not* measurement results. |
+| conditionals on measurement results           | Conditional blocks are used where the condition depends on values that depend on measurement results. |
+| dynamic qubit allocation                      | Qubits are allocated at runtime (e.g. inside loop bodies). |
+| qubit reuse                                   | Existing qubits are reset and reused at runtime. |
```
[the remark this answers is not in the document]
dime10: Thanks!

rfcs/text/0032-structured-benchmark-programs.md
```diff
@@ -0,0 +1,246 @@
...
+- Start Date: 2025-10-31
+- RFC PR: [unitaryfoundation/jeff#0032](https://github.com/unitaryfoundation/jeff/pull/0032)
+
+# Summary
+[summary]: #summary
+
+Quantum programs involving structured control flow operations such as `if` and `for` can only be represented in severely limited ways through existing quantum program formats. Qiskit does not allow dynamic qubit indexing, which is crucial for semantic `for` loops. Similarly, while OpenQASM 3 does not explicitly prohibit dynamic indexing, it leaves it open as an optional feature that many backends do not support. Jeff, on the other hand, has the ability to represent such programs, making it the ideal format for expressing a set of structured benchmark programs. These programs can then be used as a set of challenges for quantum compilers to drive the development of more advanced compilation techniques.
+
+This RFC proposes to:
+- gather and provide a list of important patterns and algorithms that contain structure and/or dynamism
```
burgholzer:
```suggestion
- gather and provide a list of important patterns and algorithms that contain structure and/or dynamism.
```
Just for consistency

rfcs/text/0032-structured-benchmark-programs.md
burgholzer: I think we should be good here. We can use inline script metadata for the Python scripts that generate benchmarks. `uv` has great support for these.
See https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies for some good documentation on that.

burgholzer (approved):
Great work @DRovara 👏🏼 
Just spotted a typo and a missing dot. Otherwise this is spot on 🎯

PR #41, opened 2026-02-10
feat: Semver versioning for the jeff format

### LINKED ISSUE #39: Add semver versioning to the spec
Start versioning the spec from `0.1.0`, so we can state compatibility in the different sub-projects and clearly indicate breaking/non-breaking changes.

aborgna-q:
Closes #39.

Adds semver version constants 
```capnp
const schemaVersionMajor: UInt32 = 0;
const schemaVersionMinor: UInt32 = 1;
const schemaVersionPatch: UInt32 = 0;
```
to the format, as well as fields in the module (We already had a `version` field, so I'm reusing it for the mayor to avoid breakage).
```capnp
struct Module {
    version @0 :UInt32;
    # The mayor version of the format.
    #
    # Forms a semver version triplet with `versionMinor` and `versionPatch`.
    versionMinor @7 :UInt32;
    # The minor version of the format.
    versionPatch @8 :UInt32;
    # The patch version of the format.
    ...
}
```

Adds code in the rust reader to validate the incoming program's version, and emit an error if not supported.

drive-by: Update the `capnp` dependency for the rust project.
drive-by: Add some docs to the start of `jeff.capnp`.


examples/python_optimization/python_optimization.ipynb
burgholzer: Jupyter diffs are always kind of hard to read, but the output here lists `versionMinor = 0`; shouldn't that be equal to one after this PR?

impl/rs/src/reader/module.rs
```diff
@@ -20,8 +20,11 @@ impl<'a> Module<'a> {
     }
 
     /// Version of the jeff protocol used in this module.
-    pub fn version(&self) -> u32 {
-        self.module.get_version()
+    pub fn version(&self) -> semver::Version {
+        let mayor = self.module.get_version() as u64;
+        let minor = self.module.get_version_minor() as u64;
+        let patch = self.module.get_version_patch() as u64;
+        semver::Version::new(mayor, minor, patch)
```
burgholzer:
```suggestion
        let major = self.module.get_version() as u64;
        let minor = self.module.get_version_minor() as u64;
        let patch = self.module.get_version_patch() as u64;
        semver::Version::new(major, minor, patch)
```

burgholzer (changes requested):
This looks good to me except for some very small corrections 😌
Hope those make sense.

examples/python_optimization/python_optimization.ipynb
burgholzer: Ah. That might really become a little cumbersome over time. Especially if people then have to touch code in languages that they may not have used before. Fine for now, but might really be worth to split off the spec from the implementations.

PR #48, opened 2026-04-02
feat: Add MBQC benchmark programs

mark-koch:
Here is my take on the MBQC benchmark set. I took a subset of the standard benchmark circuits from https://github.com/meamy/feynman and converted them into an MBQC pattern using `graphix` which I then convert to jeff.

I wasn't entirely sure about the directory structure. Is it ok to put the code to generate the benchmarks into the `_recipies` directory?


benchmarks/_recipes/structured/mbqc/generate.py
```diff
@@ -0,0 +1,180 @@
...
+                case Plane.XZ:
+                    gate = "ry"
+                    angle = -bloch.angle
+                case Plane.YZ:
+                    gate = "rx"
+                    angle = bloch.angle
+                case _:
+                    raise NotImplementedError
+            assert isinstance(angle, float | int)
+            [jeff_angle] = self.add_op(jeff.const_float(angle * math.pi, 64)).outputs
```
mark-koch: TODO: Need to check if this radians conversion is correct

impl/py/src/jeff/__init__.py
```diff
@@ -1914,10 +1914,20 @@ def qubit_free(qubit: JeffValue):
...
+    return JeffOp(
+        "float", f"const{bitwidth}", inputs, outputs, instruction_data=value
+    )
+
+
 def quantum_gate(
     name: str,
     qubits: JeffValue | list[JeffValue],
-    params: list[float] = None,
+    params: list[JeffValue] = None,
```
mark-koch, line 1920: Drive-by: Passing Python floats as params causes an error when encoding via capnproto. They should be jeff float values

burgholzer:
Thanks for working on this @mark-koch 🙏🏼 

Sorry for probably being the stupid one here, but none of these benchmarks are structured benchmarks or am I missing something?
The fact alone that they are expressed in OpenQASM 2 means that they almost cannot be.
The circuits do not even contain measurements, let alone conditionals conditioned on these measurements. I also do not see any kind of loops or branching in the files that I checked.
Could you clarify if that was intended?

mark-koch:
The original circuits are flat, but the MBQC conversion introduces measurements and classical branches conditioned on mid-circuit measurement outcomes. If you load the jeff files you should see measurements and branches

burgholzer:
> The original circuits are flat, but the MBQC conversion introduces measurements and classical branches conditioned on mid-circuit measurement outcomes. If you load the jeff files you should see measurements and branches

Ah, I figured I must have been stupid.
It's a bit unfortunate that one cannot directly inspect the jeff programs as part of the PR review, because binary files cannot be displayed.
Maybe it would be worth to also store the human-readable version in some way...

PR #50, opened 2026-04-08
fix!: Fix order of inputs to `insertIndex` and `insertSlice`

### LINKED ISSUE #42: Inconsistency between `setIndex` and `insertIndex`
I have noticed that the input order is inconsistent between `IntArrayOp.setIndex`

https://github.com/unitaryfoundation/jeff/blob/911af55c50a78e717ac7b95f2e45aeff1c37c388/impl/capnp/jeff.capnp#L944-L953

and `QuregOp.insertIndex`

https://github.com/unitaryfoundation/jeff/blob/911af55c50a78e717ac7b95f2e45aeff1c37c388/impl/capnp/jeff.capnp#L536-L547

Namely, the index is the second argument in the former case and the last in the latter.

I would propose aligning the argument order. One could also consider aligning the function names.

denialhaag:
This PR fixes the order of inputs to `insertIndex` and `insertSlice`.

Fixes #42


.pre-commit-config.yaml
```diff
@@ -59,6 +59,13 @@ repos:
       #  language: system
       #  files: \.py$
       #  pass_filenames: false
+      - id: nb-clean
+        name: nb-clean
+        description: Clean Jupyter notebooks with `nb-clean`.
+        entry: uv run nb-clean clean --preserve-cell-outputs .
+        language: system
+        files: \.ipynb$
+        pass_filenames: false
```
burgholzer, line 68: `nb-clean` has a dedicated pre-commit check (which we used in the past in the MQT: https://github.com/munich-quantum-toolkit/core/blob/ae5406d206524ff49d79a9dbe93518e823154479/.pre-commit-config.yaml#L33-L42)

pyproject.toml
```diff
@@ -3,7 +3,7 @@ members = ["impl/py"]
 
 [dependency-groups]
 dev = [{ include-group = "lint" }, { include-group = "test" }]
-lint = ["pre-commit ~=4.3", "ruff ~=0.12.10", "mypy ~=1.17"]
+lint = ["pre-commit ~=4.3", "ruff ~=0.12.10", "mypy ~=1.17", "nb-clean ~=4.0.1"]
```
burgholzer, line 6:
```suggestion
lint = ["pre-commit ~=4.3", "ruff ~=0.12.10", "mypy ~=1.17"]
```
As per the comment in the pre-commit file.

denialhaag:
I hope this PR already includes all the necessary changes to fix #42. Please let me know if something else needs to be done. 🙂 I guess the real work will be updating `jeff-mlir` and any other tools that read or write Jeff files.

In #42, I initially proposed aligning the naming of `insertIndex` and `setIndex`, but I'm no longer sure I like that idea. `insertIndex` conveys that the linear qubit can no longer be used after it is inserted, while an integer or float can still be used after an array entry is set to it. I'm open to other opinions here. 🤔

burgholzer (approved):
This looks pretty good to me, so I am pre-approving this from my side.
I wasn't fully sure whether https://github.com/unitaryfoundation/jeff/issues/42#issuecomment-4155056485 is addressed by this given there are not too many changes to the spec itself.

burgholzer:
> In https://github.com/unitaryfoundation/jeff/issues/42, I initially proposed aligning the naming of insertIndex and setIndex, but I'm no longer sure I like that idea. insertIndex conveys that the linear qubit can no longer be used after it is inserted, while an integer or float can still be used after an array entry is set to it. I'm open to other opinions here. 🤔

I can get myself behind that reasoning.

.pre-commit-config.yaml
denialhaag: I initially wanted to use the dedicated hook, but this is more in line with the other hooks in the file. For example, `ruff` is also run as a `system` hook. 🤔

burgholzer: Ah, nvm then. Should have looked a little closer at the surrounding code. Feel free to resolve the suggestions.

impl/py/src/jeff/data/jeff.capnp
dime10, line 1: Is this file checked in twice ? 😅

dime10 (approved):
Thanks @denialhaag :) Agreed about keeping insert* vs set*

impl/py/src/jeff/data/jeff.capnp
burgholzer: Glad that I'm not the only one being surprised by this. @aborgna-q explained that at some point in another PR. Apparently the python bindings need the definition in-source as part of the python package.

denialhaag: Yes, I also read that in a README or something like that.

The file is automatically kept in sync with the main one via `just update-capnp`, though.

PR #51, opened 2026-04-09
docs: Streamline capitalization and stylization of `jeff`

denialhaag:
After an offline discussion, we have decided to stylize `jeff` as `jeff` (and not jeff, Jeff, *jeff*, or *Jeff*). This PR implements this decision across (hopefully all) Markdown files.


aborgna-q (approved):

benchmarks/structured/README.md
```diff
@@ -11,7 +11,7 @@ This section tracks the current programs of benchmark program implementation and
 
 The following table lists the currently implemented benchmark programs and indicates the different formats the program is available in.
 
-| Program                                                          | Jeff |                   OpenQASM 3                     |
+| Program                                                          | `jeff` |                   OpenQASM 3                     |
```
burgholzer: You may want to autoformat the table here again as the columns no longer match up due to the additional characters.

burgholzer (approved):
Nice 🙌🏼 Just one minor nitpick from my side

benchmarks/structured/README.md
denialhaag: Good point! I have now formatted the entire file with `prettier`.

We may want to consider adding that to our `pre-commit` hooks, maybe in a separate PR, though. 🤔

burgholzer: Sounds good to me for a separate PR!
I am going to queue this for merge now.
We can always come back to this and change it again if we feel like it.

PR #52, opened 2026-04-09
feat!: Add size information to `qureg`, `intArray`, and `floatArray` types

### LINKED ISSUE #46: Add size information to `FloatArray`, `IntArray` and `Qureg` types
Currently, the `FloatAtray`, `IntArray` and `Qureg` types do not contain any information on the respective object's size. It might make sense to **optionally** store this information in the type.

I have noticed in the conversion from Jeff MLIR to MQT IR that we cannot write round-trip tests because the Jeff types do not contain size information. When allocating a tensor in MLIR it is automatically typed as, for example, `tensor<5x!qco.qubit>` (a tensor containing five qubits from our QCO dialect). Round-tripping such a type results in `tensor<?x!qco.qubit>` (a tensor containing a dynamic number of qubits) because Jeff has not preserved the size information. It is certainly possible to work around this somehow and create extract the size information from the context, but this feels somewhat wrong to me.

Could it make sense add size information to `FloatAtray`, `IntArray` and `Qureg` types? Or is such a a round-trip not necessarily intended to work?

denialhaag:
This PR adds size information to the `qureg`, `intArray`, and `floatArray` types. In all three cases, the size can still be dynamic.

Fixes #46


denialhaag:
@aborgna-q, @burgholzer, and @dime10, this should also be ready for a first review. 🙂 

(Could you add me as a collaborator so that I can directly request reviews via GitHub? 🤔)

impl/rs/src/types.rs
```diff
@@ -20,8 +20,13 @@ pub enum Type {
     ///
     /// Quantum registers are linear types.
     ///
-    /// The length of the register is not known at compile time, but fixed at runtime.
-    QubitRegister,
+    /// If `length` is `None`, the register has dynamic length.
+    /// If `Some`, the register has static compile-time length.
+    #[display("Qureg")]
```
denialhaag: It probably makes sense to add the size to `display` here (and also further below for `IntArray` and `FloatArray`).

impl/py/src/jeff/__init__.py
```diff
@@ -93,8 +93,11 @@ def __eq__(self, other):
...
 
-        if hasattr(self, "bitwidth"):
-            return self.bitwidth == other.bitwidth
+        if hasattr(self, "bitwidth") and self.bitwidth != other.bitwidth:
+            return False
+
+        if hasattr(self, "length") and self.length != other.length:
+            return False
 
         return True
```
denialhaag: Thinking about it more, the special cases in `__eq__` should probably not be defined in `JeffType` but in the specific classes themselves. 🤔

impl/capnp/jeff.capnp
```diff
@@ -338,15 +338,17 @@ struct Type {
...
+        qureg :union {
+            dynamic @1 :Void;
+            static @2 :Length;
+        }
         # Quantum registers.
         #
         # A quantum register is an array of slots that can hold qubits.
         # Slots of a quantum register can either be empty or filled with a qubit.
         # Quantum registers are linear types.
-        # The length of the register is not known at compile time.
```
burgholzer, line 347: This probably deserves additional documentation explaining the two different types of registers

impl/capnp/jeff.capnp
```diff
@@ -355,22 +357,28 @@ struct Type {
...
+            bitwidth @4 :Bitwidth;
+            length :union {
+                dynamic @5 :Void;
+                static @6 :Length;
+            }
+        }
         # Integer arrays.
         #
-        # The length of the array is not known at compile time.
-        #
```
burgholzer, line 362: Probably also deserves a bit more documentation.

impl/capnp/jeff.capnp
```diff
@@ -355,22 +357,28 @@ struct Type {
...
+        floatArray :group {
+            precision @8 :FloatPrecision;
+            length :union {
+                dynamic @9 :Void;
+                static @10 :Length;
+            }
+        }
         # Floating point number arrays.
-        #
-        # The length of the array is not known at compile time.
```
burgholzer, line 373: Same as above

impl/capnp/jeff.capnp
burgholzer, line 1: This change will also definitely require a version bump of the spec.

impl/py/src/jeff/__init__.py
burgholzer: Yeah; given how only a few types even have these attributes, these are dead checks for most operations. I am in favor of moving them to the individual operations.

impl/py/src/jeff/__init__.py
```diff
@@ -116,11 +119,31 @@ def __str__(self) -> str:
...
 
+    # static fields
+
+    @property
+    def length(self) -> int | None:
+        if self._length is not _Empty:
+            return self._length
+
+        if self._raw_data.floatArray.length.which == "static":
+            return self._raw_data.floatArray.length.static
```
burgholzer: Probably just me being stupid, but what is `floatArray` doing in the `QuregType` code?

impl/rs/src/types.rs
burgholzer: Seems reasonable, yeah.

burgholzer (changes requested):
This generally looks good to me. I only have some minor remarks here and there.

impl/rs/src/types.rs
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/52/commits/2565b0ba8161409929e948f65aa7c2278fa88270.

impl/py/src/jeff/__init__.py
denialhaag: No, I just copy-pasted blindly... This automatically came up in https://github.com/unitaryfoundation/jeff/pull/52/commits/2565b0ba8161409929e948f65aa7c2278fa88270. Thanks for catching that!

impl/py/src/jeff/__init__.py
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/52/commits/2c0431c6854305a4133e14c6d3765d1a58f58b21.

impl/capnp/jeff.capnp
denialhaag: Good point! Done.

impl/capnp/jeff.capnp
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/52/commits/03a09f9b8ce6ab7f7f12f885568249a38139aa67.

impl/capnp/jeff.capnp
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/52/commits/03a09f9b8ce6ab7f7f12f885568249a38139aa67.

impl/capnp/jeff.capnp
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/52/commits/03a09f9b8ce6ab7f7f12f885568249a38139aa67.

impl/capnp/jeff.capnp
```diff
@@ -338,15 +338,20 @@ struct Type {
...
+        }
         # Quantum registers.
         #
         # A quantum register is an array of slots that can hold qubits.
         # Slots of a quantum register can either be empty or filled with a qubit.
         # Quantum registers are linear types.
-        # The length of the register is not known at compile time.
+        #
+        # The quantum register can have a static compile-time length.
+        # A dynamic length can be selected by setting the corresponding flag.
```
burgholzer: maybe something along the lines of
> The length of the register can be a compile-time constant (static) or dynamic.

is a better fit?
seems to be more in line with the previous docstring.

impl/py/src/jeff/__init__.py
```diff
@@ -112,19 +101,51 @@ def _refresh(self, new_data: schema.Type.Builder):
...
+        if self._raw_data.qureg.which == "static":
+            return self._raw_data.qureg.static
+        else:
+            return None
+
     # Python integration
 
     def __str__(self) -> str:
-        return "qureg"
+        return f"qureg[{self.length if self.length is not None else ''}]"
```
burgholzer: would either

```
        return f"qureg{f"[self.length]" if self.length is not None else ''}]"
```
(basically rendering dynamic register just like before)
or

```
        return f"qureg[{self.length if self.length is not None else '?'}]"
```
(which closely mirrors MLIR)

be better here?
The first solution would definitely minimize the diff here, while the second one would be more explizit.

Also applies to the other `__str__` methods.

impl/rs/src/types.rs
```diff
@@ -20,8 +20,13 @@ pub enum Type {
     ///
     /// Quantum registers are linear types.
     ///
-    /// The length of the register is not known at compile time, but fixed at runtime.
-    QubitRegister,
+    /// If `length` is `None`, the register has dynamic length.
+    /// If `Some`, the register has static compile-time length.
+    #[display("Qureg{}", length.map_or("".to_string(), |l| format!("[{}]", l)))]
```
burgholzer: Just so we do not forget. The formatting question also applies here.

burgholzer (changes requested):
Thanks Daniel!
I have a couple of further smaller comments on this.

And one more general one: so far, the new static sizes aren't really used anywhere. Shouldn't we also use them where possible?
I am a bit surprised that some of the examples that define the size as an integer constant do not carry the static type information now.

impl/capnp/jeff.capnp
denialhaag: I like it! Done.

impl/py/src/jeff/__init__.py
denialhaag: Done. I went with `[?]` for dynamic sizes.

impl/rs/src/types.rs
denialhaag: Done. I went with `[?]` for dynamic sizes.

denialhaag:
> And one more general one: so far, the new static sizes aren't really used anywhere. Shouldn't we also use them where possible? I am a bit surprised that some of the examples that define the size as an integer constant do not carry the static type information now.

Good point! I have improved this a bit in https://github.com/unitaryfoundation/jeff/pull/52/commits/8023350d1b5e1798cf4849753069e33d250c275e. I don't really feel comfortable editing the Catalyst examples in this case because a translation from Catalyst could not reasonably create qubit registers with a static size. 🤔

dime10:
> (Could you add me as a collaborator so that I can directly request reviews via GitHub? 🤔)

@bachase Would you mind? 🙂

burgholzer (approved):
I am happy with this 🙌🏼 Thanks for incorporating the change requests so quickly!
I'll leave it to someone else to perform an additional review!

impl/capnp/jeff.capnp
```diff
@@ -338,15 +338,19 @@ struct Type {
...
         #
         # A quantum register is an array of slots that can hold qubits.
         # Slots of a quantum register can either be empty or filled with a qubit.
         # Quantum registers are linear types.
-        # The length of the register is not known at compile time.
+        #
+        # The length of the register can be a compile-time constant (static) or dynamic.
 
-        int @2 :Bitwidth;
+        int @3 :Bitwidth;
```
aborgna-q: When possible we should preserve the item labels. This makes it so we can load older jeffs with the new definition.

In the case of `intArray @3 :Bitwidth;` we should reuse the label for the bitwidth,
```capnp
intArray :group {
    bitwidth @3 :Bitwidth;
    ...
```

aborgna-q (approved):
Nice!

impl/capnp/jeff.capnp
dime10: good point, I guess it's the static field above `static @2 :Length;` that should have a new number that isn't taken yet

burgholzer: That was something I was wondering about too. I know from my limited experience with protobuf that it is kind of best practice to always use new identifiers instead of reindexing old ones.
Does this also apply for cap'n'proto?

dime10: I believe so, this has more information on it: https://capnproto.org/language.html#evolving-your-protocol

dime10: > Remember that the purpose of the numbers is to indicate the evolution order of the struct. The system needs to know when the union fields were declared relative to the non-union fields.
>
> New fields, enumerants, and methods may be added to structs, enums, and interfaces, respectively, as long as each new member’s number is larger than all previous members. Similarly, new fields may be added to existing groups and unions.
>
> Members can be re-arranged in the source code, so long as their numbers stay the same.
>
> You cannot change a field, method, or enumerant’s number.

denialhaag: That makes a lot of sense; thanks for pointing that out!

https://github.com/unitaryfoundation/jeff/pull/52/commits/afe2a0c03835528c8e89ec9e30c5dc9311607e7a should fix this.

`@1` was `qureg` before. Now, it's `qureg.dynamic`. Does that make sense, or should we no longer use `@1`? 🤔

dime10: I think so:
> A field can be moved into a group or a union, as long as the group/union and all other fields within it are new. In other words, a field can be replaced with a group or union containing an equivalent field and some new fields.

burgholzer (approved):
Nice. I think this should be good to go now!

PR #62, opened 2026-06-03
feat: add jeff verifier tool

### LINKED ISSUE #57: Add verification tool

huhaha120:
## Summary

- Add a Python verifier under `tools/verifier/` with a CLI entry point.
- Validate module structure, value table references, use-before-definition, nested region isolation, instruction input/output signatures, integer bitwidth / float precision consistency, and qubit / qureg linearity.
- Add focused positive and negative tests for the requested verifier checks, plus README usage notes.

Closes #57

## Tests

- `uv run ruff format --check`
- `uv run ruff check`
- `uv run pytest -q`
- `uv run python -m tools.verifier examples/*/*.jeff`

## AI Assistance Disclosure

This contribution was AI-assisted. I reviewed the verifier behavior against the issue requirements, checked the CLI and examples manually, and verified the implementation locally with the tests listed above.

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -1,0 +1,1 @@
+"""Repository-local tooling."""
--- a/tools/verifier/README.md
+++ b/tools/verifier/README.md
@@ -1,0 +1,23 @@
+# jeff verifier
+
+This tool validates encoded `.jeff` modules beyond the structural guarantees provided
+by the Cap'n Proto schema.
+
+Run it from the repository root:
+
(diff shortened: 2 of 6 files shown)


denialhaag:
Thanks for your interest in contributing to `jeff` in the context of unitaryHACK. Unfortunately, we will have to reject this contribution due to several violations of the unitaryHACK rules. We are happy to review and accept a rules-compliant contribution.

PR #65, opened 2026-06-07
Add jeff program verification helpers

### LINKED ISSUE #57: Add verification tool

Forage409:
Closes #57.

This adds a small verification module to the Rust reader API:

- `verify(&impl ReadJeff)` and `verify_module(Module)` entry points
- schema major-version and entrypoint validation
- function signature value resolution checks
- dataflow checks for values used before they are defined
- nested structured-control-flow region checks
- function-call target validation
- tests covering the existing example programs plus an invalid entrypoint case

Tested with:

```text
CARGO_HTTP_CHECK_REVOKE=false cargo test -q
```

Disclosure: I used AI assistance while implementing and checking this change.

### DIFF
--- a/impl/rs/src/reader.rs
+++ b/impl/rs/src/reader.rs
@@ -9,6 +9,7 @@ mod op;
 mod region;
 mod string_table;
 pub mod value;
+pub mod verify;
 
 pub mod optype;
 
@@ -17,6 +18,7 @@ pub use metadata::{HasMetadata, Metadata};
 pub use module::Module;
 pub use op::Operation;
 pub use region::Region;
+pub use verify::{verify, verify_module, VerificationError};
 
 use derive_more::derive::{Display, Error, From};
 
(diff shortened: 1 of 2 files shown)


denialhaag (changes requested):
Thanks for your interest in contributing to `jeff`, @Forage409! 🙂 

Before I can review this, please add more tests (at least one negative test per bullet point in #57, but more is always better). That way, we both get a sense of what your implementation already covers. If you need help with generating `jeff` programs, see [`encode_examples.sh`](https://github.com/unitaryfoundation/jeff/blob/main/examples/encode_examples.sh) or [`generate_test_cases.py`](https://github.com/PennyLaneAI/jeff-mlir/blob/main/unittests/inputs/generate_test_cases.py).

denialhaag:
We have decided to go with #68 to resolve #57, so I'll close this PR.

Thanks again for your interest in contributing to `jeff`, @Forage409! 🙂

PR #66, opened 2026-06-07
feat: Add verification tool for jeff programs

### LINKED ISSUE #57: Add verification tool

kawacukennedy:
Closes #57.

Adds verification tool at `tools/verifier/` that checks all 7 requirements from the issue:

1. **Required module attributes** — entrypoint index bounds check
2. **SSA validity** — ensures all value references are within the value table
3. **Def-before-use** — catches out-of-bounds value IDs in inputs/outputs
4. **Type consistency** — validates input/output types match all operations (QubitOp, QubitRegisterOp, IntOp, IntArrayOp, FloatOp, FloatArrayOp, ControlFlowOp)
5. **Int/float operand consistency** — same-bitwidth and same-precision checks for binary arithmetic operations
6. **Qubit linearity** — tracks qubit/qureg consumption across operations and nested regions
7. **Region isolation** — ensures nested regions only reference values from region sources or definitions within the region

### Implementation
- Rust crate at `tools/verifier/` using the reader API (`jeff::reader`)
- CLI binary with `clap` for command-line verification
- 5 modular check passes: `attributes`, `ssa`, `types`, `linearity`, `isolation`
- Recursive handling of nested control-flow regions (For, While, DoWhile, Switch)
- 9 integration tests: 3 positive (existing example programs) + 6 negative (invalid fixtures)
- Tested via `cargo test -p jeff-verify`

### CI integration
- `tools/verifier/**` added to the Rust change filter so CI triggers on verifier changes
- Workspace `Cargo.toml` updated to include `tools/verifier`

### Notes
- Uses `matches!` for array type length wildcards (avoids false positives on intArray/FloatArray length mismatches)
- Valid examples `qubits`, `entangled_qs`, and `entangled_calls` all pass cleanly


kawacukennedy:
@denialhaag @burgholzer — this implements all 7 checks from issue #57. Could you take a look?

tools/verifier/src/checks/attributes.rs
```diff
@@ -0,0 +1,28 @@
...
+            errors.push(CheckError {
+                check_name: self.name(),
+                message: format!(
+                    "Entrypoint index {entry_id} is out of bounds: module has {func_count} function(s)."
+                ),
+            });
+        }
+
+        errors
+    }
```
denialhaag, line 35: This should probably raise an error if `entrypoint` is not specified (see, e.g., [`entangled_qs.txt`](https://github.com/unitaryfoundation/jeff/blob/main/examples/entangled_qs/entangled_qs.txt))

tools/verifier/src/checks/isolation.rs
denialhaag, line 1: It looks like this is not tested.

tools/verifier/src/main.rs
```diff
@@ -0,0 +1,56 @@
+//! Verification tool for jeff programs.
+//!
+//! Checks a jeff file for common issues including:
+//! - Required module attributes (entrypoint, tool name/version)
+//! - SSA validity (all values defined before use)
+//! - Type consistency for operations
+//! - Qubit linearity (qubits consumed at most once)
```
denialhaag, line 7: Qubits must be used exactly once. Every qubit must be either destructively measured or freed. The respective check should be adapted and tested accordingly.

tools/verifier/src/main.rs
```diff
@@ -0,0 +1,56 @@
+//! Verification tool for jeff programs.
+//!
+//! Checks a jeff file for common issues including:
+//! - Required module attributes (entrypoint, tool name/version)
```
denialhaag: Strictly speaking, the tool name and version are not required attributes.

tools/verifier/src/checks/types.rs
denialhaag, line 1: The checks here look pretty good already, but it would be nice to improve the test coverage beyond the `add` operations.

denialhaag (changes requested):
Thanks for also working on the verifier, @kawacukennedy! 🙂 

I had an initial look at your implementation, and you can find some feedback below. This is already going in the right direction!

kawacukennedy:
Hi @denialhaag — f9eb7a4 addresses your review: entrypoint fix, linearity leak detection, more tests. Please take another look.

PR #68, opened 2026-06-10
feat: add verification tool

### LINKED ISSUE #57: Add verification tool

orisus42:
This PR closes #57 

The verifier is at `tools/verifier`. Its public API is one function:
`pub fn verify_module(module: Module<'_>) -> Vec<VerificationError>`

Callers can pass the decoded module from `Jeff::read(...).module()` and get back a list of every verification error found. An empty list means the module is valid.

## Architecture
The verifier iterates over all functions in the module. For each function definition it runs sequential passes. Declarations (functions without a body) are skipped since they have no code to verify.

All passes append to the same shared `Vec<VerificationError>`. The verifier is non-fatal, such that it collects every error it finds rather than stopping at the first one. This lets it show the user all the invalid structures in the program.

The current passes check the following requirements as stated in the issue:
- [x] Are all required module attributes (e.g., version and entrypoint) set?
- [x] Is a value used before it is defined?
- [x] Are all values used in a function's body defined in its values list?
- [x] Are all input and output values of the expected type?
- [x] Are the input and output values of a given integer or float operation (e.g., add or max) of the same type (i.e., the same bitwidth or float precision)?
- [x] Is linearity respected, i.e., are all qubit and qureg values used exactly once?
- [x] Are all regions isolated from above?

### Pass 1 (Module Attributes)
**Version Check** - A value of `0.0.0` means the version was not set.
**Entrypoint** - `module.entrypoint_id()` must refer to an index that exists in the function list AND that function must be a definition, not a declaration. Anything else raises a verification error.

### Pass 2 (Value Checks)
**Out-of-bounds** - The `reader` yields `ReadError:ValueOutOfBounds` which is piped to a verification error.
**Use-before-defined** - This checks for programs where an operator consumes a value that is produced later (or never produced at all).
**Linearity** - Qubits and qubit registers are linear types: they must be produced exactly once and consumed exactly once. The analysis module walks the entire region tree and increments per-value producers and consumers counters:
    - A region source counts as one producer.
    - An op output counts as one producer.
    - A region target counts as one consumer.
    - An op input counts as one consumer.

### Pass 3 (Type Checks)
For each operation the verifier reads the types of its actual inputs and outputs from the value table, then checks them against the operation's expected signature. The pass recurses into all nested regions so every operation in the program is covered.

### Pass 4 (Region Isolation)
It follows MLIR's `IsolatedFromAbove` semantics for structured control flow regions: a nested region cannot directly reference a value from an outer scope. Values must be explicitly passed in via the region's sources list.

Isolation violations no longer trigger `UsedBeforeDefined`.
`UsedBeforeDefined` is an **ordering** check which checks if this value was produced before it was used, within this scope. Outer-scope values ARE produced before the nested region runs, so the answer is yes.
`IsolationViolation` is a **scoping** check which asks whether it captured something from outside without going through sources. 

## Future changes
- Inter-function checks can be implemented
- Linearity is checked globally, it can be made branch-sensitive
- Arity of region inputs and outputs can be checked
- Similarly, arity of measurements can also be checked
- Verification errors can be made to point to the exact instruction that fails the check
- An option to stop the verifier once the first verification error occurs would be useful for debugging

## Tests
Some tests that cover all the verification error codes were added. Other tests can be generated on demand using a Python or Bash script in future updates.

### AI Disclosure:
The architecture and design process was completely human, GPT-5.5 was used to assist me in understanding concepts from the repository and verify my thoughts for the implementation. GPT-5.5 and Claude Sonnet 4.6 was used in writing code and generating the large set of test cases. Each section of code for the `verifier` was generated only after scrutiny  by me. Most of the test cases generated were verified by me, and running the verifier on the complete set of test cases passed as expected. None of the above PR description was generated by any LLM.


orisus42:
Some CI checks are failing due to me removing the tests at the end and replacing them with the scripts. I'll update the PR so that those tests aren't run by default until the generator scripts are run.

tools/verifier/src/errors.rs
```diff
@@ -0,0 +1,148 @@
...
+        value_id: u32,
+    },
+
+    /// A linear value (qubit or qureg) is produced by more than one operation.
+    LinearValueProducedMultipleTimes {
+        /// The value produced multiple times.
+        value_id: u32,
+        /// The number of producing operations.
+        producers: u32,
+    },
```
denialhaag: I think this error can be generalized. See [this section in the spec](https://github.com/unitaryfoundation/jeff/blob/b98791ed02d75aae8517584fde3943006ee9f64a/docs/spec.md#L15-L16):

> `jeff` uses *value semantics*: values are immutable, and operations produce new values rather than modifying existing ones.

The two errors below are exclusive to qubits and quregs, however.

tools/verifier/src/passes/module_attributes.rs
```diff
@@ -0,0 +1,20 @@
...
+use jeff::reader::{Function, Module};
+
+use crate::VerificationError;
+
+/// Verify that the module has a non-zero version and a valid entrypoint.
+pub fn verify_module_attributes(module: Module<'_>, errors: &mut Vec<VerificationError>) {
+    let v = module.version();
+    if v.major == 0 && v.minor == 0 && v.patch == 0 {
+        errors.push(VerificationError::MissingVersion);
+    }
```
denialhaag: I'm wondering whether this should enforce something slightly stricter. Most likely, a given version of the verifier can only be used on `jeff` programs written for a specific version of the spec.

That said, this may be too hypothetical at this point and can be addressed when the issue arises in the future. Furthermore, if a version check is needed, it should probably already occur in `lib.rs`, and the verifier should immediately exit, stating that it is incompatible with the `jeff` program.

tools/verifier/src/passes/value_checks.rs
```diff
@@ -0,0 +1,127 @@
...
+        Ok(stats) => {
+            for (id, value) in values.iter() {
+                if is_linear(value.ty()) {
+                    let stat = &stats[id as usize];
+                    if stat.producers > 1 {
+                        errors.push(VerificationError::LinearValueProducedMultipleTimes {
+                            value_id: id,
+                            producers: stat.producers,
+                        });
+                    }
```
denialhaag: As stated already in my comment above, this should be enforced for any value.

tools/verifier/src/passes/isolation.rs
```diff
@@ -0,0 +1,84 @@
...
+        .map(|v| v.id())
+        .collect();
+
+    for op in region.operations() {
+        for input in op.inputs().filter_map(|r| r.ok()) {
+            if outer_values.contains(&input.id()) && !locally_defined.contains(&input.id()) {
+                errors.push(VerificationError::IsolationViolation {
+                    value_id: input.id(),
+                });
+            }
```
denialhaag, line 32: Correct me if I'm wrong, but the two checks seem to be of a different nature.

1. If `outer_values.contains(&input.id())`, it's definitely an `IsolationViolation`.
2. If `!locally_defined.contains(&input.id())`, it's more of a `UsedBeforeDefined` error, right?

tools/verifier/src/passes/isolation.rs
```diff
@@ -0,0 +1,84 @@
...
+            check_cf_isolation(cf_op.as_ref(), &new_outer, errors);
+        }
+    }
+
+    for target in region.targets().filter_map(|r| r.ok()) {
+        if outer_values.contains(&target.id()) && !locally_defined.contains(&target.id()) {
+            errors.push(VerificationError::IsolationViolation {
+                value_id: target.id(),
+            });
+        }
```
denialhaag, line 54: Same comment here.

tools/verifier/src/passes/type_checks.rs
```diff
@@ -0,0 +1,579 @@
...
+                }
+            }
+        }
+        _ => {}
+    }
+}
+
+fn is_qureg(ty: &Type) -> bool {
+    matches!(ty, Type::QubitRegister { .. })
+}
```
denialhaag: It might make sense to also define similar `is_int()`, `is_float()`, and so on functions. I'll let you decide if that makes the code cleaner!

tools/verifier/tests/negative/encode/bounds/region_source_oob.txt
```diff
@@ -0,0 +1,20 @@
...
+        body = (
+          sources = [5],
+          targets = [],
+          operations = [
+            ( inputs = [],
+              outputs = [0],
+              instruction = (int = (const32 = 10)) ) ] ),
+        values = [
+          (type = (int = 32)),
+          (type = (int = 32)) ] ) ) ],
```
denialhaag: This should also already "work" with just a single value, right? 🤔

tools/verifier/tests/negative/encode/bounds/region_target_oob.txt
```diff
@@ -0,0 +1,20 @@
...
+        body = (
+          sources = [],
+          targets = [5],
+          operations = [
+            ( inputs = [],
+              outputs = [0],
+              instruction = (int = (const32 = 10)) ) ] ),
+        values = [
+          (type = (int = 32)),
+          (type = (int = 32)) ] ) ) ],
```
denialhaag: Same question here.

denialhaag (changes requested):
Thanks a lot for your contribution, @orisus42! This already looks really good to me! 🙂 

I left some minor comments; you can find them below.

Regarding the testing, I'd be completely fine if you commit the `.jeff` files for now. Don't worry if it's too much of a hassle to set up a workflow that generates the `.jeff` files on demand. We can easily do that in a follow-up!

tools/verifier/src/passes/module_attributes.rs
orisus42: It would make sense for the version check to actually confirm whether the right version is being used, however for this implementation what I found was that if the version is not set explicitly then it defaults to `0.0.0` which obviously isn't valid, hence that check suffices for an unset version check. You're right however, this check can be upgraded into two unique errors, `MissingVersion` for when the version is `0.0.0` and `IncompatibleVersion` for when it is anything different (although if a version is not set in that case then it would trigger both the errors, but it can be made more explicit by making sure `IncompatibleVersion` does not trigger on `0.0.0` which serves as an indicator of an unset version).

orisus42:
I will simplify some tests further and combine any tests that I can! Will commit the `.jeff` files with their corresponding `.txt` after cleaning the code up a bit and doing the changes as recommended.

tools/verifier/src/passes/module_attributes.rs
denialhaag: I agree that this should be treated as two separate errors. I would definitely keep `MissingVersion` as is and maybe add a separate `IncompatibleVersion` error.

tools/verifier/src/errors.rs
orisus42: You're right! I have now generalized it to _any_ value, not just linear ones (qubits and quregs).

tools/verifier/src/passes/isolation.rs
orisus42: I have now removed the second check! It seemed that I left it there when there was ongoing deliberation about how value ordering works. Removing it is completely safe, of course, as this error is already being caught by `check_region_ordering` to return `UsedBeforeDefined`.

denialhaag: Makes sense! Perfect!

tools/verifier/src/passes/module_attributes.rs
denialhaag: I like how you implemented this now! 👍

tools/verifier/tests/negative/incompatible_version.txt
denialhaag, line 1: Maybe it makes sense to commit the Bash script that encodes the `.txt` files to `.jeff` files again. This makes it easier to modify and/or add tests in the future.

denialhaag (approved):
Thanks a lot for addressing all of my comments! 🙂

Besides the one comment below, this looks good to me now. I'll approve already and leave the final review to @aborgna-q.

tools/verifier/src/passes/module_attributes.rs
orisus42: I would like to mention that the `reader` has its own version check completely separate from the verifier's. Before any verification can happen, `reader` validates the version against its own supported range (for now it means that if the version is above `0.2.x` it returns a `Err(JeffError::VersionTooNew)` and never gives a `Module` to pass to the verifier as it returns right there. Currently for the incompatible version test it uses the version `0.1.0` which doesn't cause this panic but it might be informational to document this behavior and perhaps come up with an implementation later that kind of deals with this better.

When I tried using the version `0.3.0`, our test calls `.expect("failed to parse jeff file")` on that result, which turns the `Err` into a panic. There might also be better ways for the `verifier` to have compatible versioning with `jeff` programs, might need to look into this.

denialhaag: Yeah, I guess the `verifier` could be compatible with a range of versions, for example. I'm personally fine with the current implementation. We can adjust this when the issue comes up. 🤔

denialhaag: Down the line, we could also define another function that accepts a path to a `.jeff` file. This function could catch the `VersionTooNew` error during reading and then return a "nicer" error. We can add this function when we make the verifier a CLI tool in a follow-up. 😌

tools/verifier/src/passes/type_checks.rs
```diff
@@ -0,0 +1,413 @@
...
+    errors: &mut Vec<VerificationError>,
+) {
+    match qubit_op {
+        QubitOp::Alloc => {
+            if !inputs.is_empty() {
+                errors.push(VerificationError::InvalidInputType { operation: "Alloc" });
+            }
+            expect_output(outputs, 0, is_qubit, "Alloc", errors);
+        }
+        QubitOp::Free | QubitOp::FreeZero | QubitOp::Reset => {
```
aborgna-q: Reset returns a qubit

tools/verifier/src/passes/type_checks.rs
aborgna-q, line 1: We are checking that the i/o matches the types, but we should also validate that there are no extra inputs/outputs.

aborgna-q, line 1: It may have been cleaner to define a validation trait and implement it for each variant, but no need to change that for this PR.

tools/verifier/src/passes/type_checks.rs
```diff
@@ -0,0 +1,413 @@
...
+
+        match op.op_type() {
+            OpType::IntOp(int_op) => check_int_op(int_op, &inputs, &outputs, errors),
+            OpType::FloatOp(float_op) => check_float_op(float_op, &inputs, &outputs, errors),
+            OpType::QubitOp(qubit_op) => check_qubit_op(qubit_op, &inputs, &outputs, errors),
+            OpType::QubitRegisterOp(qureg_op) => {
+                check_qureg_op(qureg_op, &inputs, &outputs, errors);
+            }
+            OpType::ControlFlowOp(cf_op) => check_cf_region_types(cf_op.as_ref(), errors),
+            _ => {}
```
aborgna-q: We're ignoring multiple variant here.

tools/verifier/src/passes/type_checks.rs
orisus42: You're right! Fixing this now.

tools/verifier/src/passes/type_checks.rs
orisus42: This can be done, I am adding I/O arity checks to all operations.

tools/verifier/src/passes/type_checks.rs
denialhaag: Ohh, yeah, I must have missed this... `IntArrayOp` and `FloatArrayOp` are missing. It's important to check that, e.g., the bitwidth of the output of `IntArrayOp.getIndex` must coincide with the bitwidth of the underlying `intArray`.

orisus42: Added `IntArrayOp` and `FloatArrayOp` checks (arity and type). For `getIndex`, `setIndex` and `create` the bitwidths coincide with the underlying array bitdwitdths. Refactored `type_checks.rs` pass for slightly cleaner code.

I have also fixed two examples from the `examples/` directory because the verifier was using those examples as positive tests and there was an issue with `setIndex` in two of the examples which caused the verifier to raise errors on the unmatching bitwidths.

One operator is left to implement which is `FuncOp` but all it does is call another function using the function's index, I haven't implemented this because currently the `TypeChecks` pass only receives the `Region` and to type check the signature (and the arity) for an inter-function call it would require that I thread the top level `Module` through the pass and needs more infrastructure (I'm guessing this type check would require a lookup through some kind of table that stores function signatures and indexes them by their `func_idx`, making sure that the called index is not out of bounds). I thought that this might require more significant refactoring (like a seperate analytic pass which builds this signature table) and if we decide to go through that then it would also mean that we could further simplify the `TypeChecks` pass.

orisus42: Would also like to request any maintainer to assign me to this issue before merging as mentioned by the `unitaryHACKS` organizing team that a bot checks assigned members to collect information about the bounties.

orisus42:
Hi @aborgna-q ! It's been a few days, just wanted to make sure if you could look into this once more and let me know your thoughts. Sorry for tagging after already requesting a review, but it would be awesome if we can have the `verifier` ready soon as the `unitaryHACKS` event has ended now!

tools/verifier/src/passes/type_checks.rs
aborgna-q: Ok for skipping `FuncOp`, but please add an explicit case with a comment explaining what we are not checking, and why.

```rs
            // ...
            OpType::FuncOp(_) => {}
            _ => {panic!("Unknown optype")}
```

tools/verifier/src/passes/type_checks.rs
```diff
@@ -0,0 +1,735 @@
...
+            OpType::FloatArrayOp(float_array_op) => {
+                check_float_array_op(float_array_op, &inputs, &outputs, errors);
+            }
+            _ => {}
+        }
+    }
+}
+
+fn check_cf_region_types(cf_op: &ControlFlowOp<'_>, errors: &mut Vec<VerificationError>) {
+    match cf_op {
```
aborgna-q, line 50:
```suggestion
    // TODO: Verify that SCF node input/output types. 
    match cf_op {
```

aborgna-q (commented):
Some minor last comments

> Would also like to request any maintainer to assign me to this issue before merging as mentioned by the unitaryHACKS organizing team that a bot checks assigned members to collect information about the bounties.

We'll assign you to the issue once the PR is merged. The bot will pick your name from there.

tools/verifier/src/passes/type_checks.rs
orisus42: I have added the comments and the case if the verifier encounters a new `OpType` from a future `jeff` update.

aborgna-q (approved):

aborgna-q:
Thanks!
CI is failing due to some unrelated dependency update. I'll merge the fix in #70 first and then update this branch.

denialhaag:
Alright, I'll go ahead and merge this now! @orisus42, thanks a lot for your contribution! @aborgna-q, thanks a lot for the additional review! 🙂

PR #72, opened 2026-06-22
fix!: Fix linearity of while operation

### LINKED ISSUE #4: While, DoWhile, and linear states
The `While` and `DoWhile` definitions contain to regions internally;
- A loop body with types `*state -> *state`
- A `condition` statement with types `*state -> bool` that decides when to stop the evaluation.

This is problematic when the state is linear; calling the condition forces us to discard the values.

The straightforward solution to this is to change the `condition` type to `*state -> bool, *state`.

---

For an alternative definition consider MLIR's [`scf.while`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop). There we are able to use different types for the inputs and outputs of the loops.
That corresponds roughly to a _jeff_ `While` where the regions have type:
- `while: *in_state -> *out_state`
- `condition: *in_state -> bool, *out_state`
- `body: *out_state -> *in_state`

Are we happy with the single-state?

denialhaag:
This PR refactors the `while` operation, heavily inspired by [`mlir::scf::WhileOp`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop).

The new `while` operation has two regions: `before` and `after`. Both regions can arbitrarily modify the state. The `before` region evaluates a condition (e.g., by measuring a qubit) and returns the condition in addition to any values that should be passed to the `after` region. If the condition is true, the `after` region executes before yielding values back to the `before` region. Once the condition is false, the `before` region returns the output values of the `while` operation.

As this implementation covers the functionality of a do-while loop, the `doWhile` operation is removed as part of this PR.

Fixes #4


impl/capnp/jeff.capnp
```diff
@@ -1434,61 +1434,31 @@ struct ScfOp {
...
-            # The region must have the signature `(... state) -> (... state)`.
-            # The outputs are passed as inputs to the condition region for the current iteration.
-
-            condition @6 :Region;
-            # The condition region that determines whether to continue looping.
-            #
-            # The region must have the signature `(... state) -> (int(1))`.
-            # The output is the condition result - true to continue, false to stop.
-            # The condition can only evaluate the state, not modify it.
+            # The region must have the signature `(... outputs) -> (... inputs)`.
```
denialhaag, line 1461: The diff could not possibly be harder to read, but we should align on the documentation here.

Once we have done so, I can copy over the result to `control_flow.rs` and `__init__.py`.

denialhaag: To make it a bit easier:

```
while :group {
    # While loop.
    #
    # The `before` region is executed at least once.
    # If the condition is true, the `after` region is executed.
    # The loop maintains a state consisting of any number of values.
    # The `before` region receives the state from the `after` region,
    # or the initial state for the first iteration.
    # The `after` region receives the state from the `before` region.
    # When the loop finishes, the final state is returned by the `before` region.
    #
    # Inputs:
    # - `... inputs`: Any number of values that are passed to the `before` region.
    #
    # Outputs:
    # - `... outputs`: Any number of values that are returned from the `before` region.

    before @3 :Region;
    # The region that evaluates whether the condition is met.
    #
    # The region must have the signature `(... inputs) -> (int(1), ... outputs)`.
    # If the condition is met, the `after` region is executed with the outputs.
    # If the condition is not met, the operation returns the outputs.

    after @4 :Region;
    # The region that is executed if the condition is true.
    #
    # The region must have the signature `(... outputs) -> (... inputs)`.
}
```

denialhaag: For reference, here's also the explanation of the changes from the PR description:

> This PR refactors the `while` operation, heavily inspired by [`mlir::scf::WhileOp`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop).
> 
> The new `while` operation has two regions: `before` and `after`. Both regions can arbitrarily modify the state. The `before` region evaluates a condition (e.g., by measuring a qubit) and returns the condition in addition to any values that should be passed to the `after` region. If the condition is true, the `after` region executes before yielding values back to the `before` region. Once the condition is false, the `before` region returns the output values of the `while` operation.

aborgna-q (approved):
Docs seem fine to me, and the validation update looks correct 👍

impl/rs/src/jeff.rs
```diff
@@ -41,7 +41,7 @@ impl<'a> Jeff<'a> {
     /// The minimum version of the jeff format supported by this reader.
     pub const MIN_COMPATIBLE_VERSION: semver::Version = semver::Version::new(0, 0, 0);
     /// The maximum version of the jeff format supported by this reader.
-    pub const MAX_COMPATIBLE_VERSION: semver::Version = semver::Version::new(0, 2, u64::MAX);
+    pub const MAX_COMPATIBLE_VERSION: semver::Version = semver::Version::new(0, 3, u64::MAX);
```
dime10: Will it keep supporting previous versions (e.g. the doWhile op)?

dime10 (approved):
Thanks @denialhaag!

impl/rs/src/jeff.rs
denialhaag: You raise a good point! I think we should also raise the `MIN_COMPATIBLE_VERSION`. I guess we'll release this in the coming days as part of `v0.3.0`. The newly added verifier will then be only compatible with `v0.3.0`.

burgholzer (approved):
Also LGTM! Once everyone's comments are addressed, this should be ready to go in 🚀

impl/rs/src/jeff.rs
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/72/commits/5de416aa3719edd0027cfed2d79bf6280c22722b!

denialhaag:
I have updated the spec in https://github.com/unitaryfoundation/jeff/pull/72/commits/ef9dadc73c33953c0cbeb624d4e8f75dbe103fae and the `MIN_COMPATIBLE_VERSION` in https://github.com/unitaryfoundation/jeff/pull/72/commits/5de416aa3719edd0027cfed2d79bf6280c22722b. The latter change required a slight change to the verifier, but I think it makes sense.

From my side, this would be ready to go now!

burgholzer (approved):
LGTM. Let' get this in. We can always address any regressions or oversights in follow-ups 🚀

PR #73, opened 2026-06-24
docs: Set up `rumdl`

denialhaag:
As the `jeff` repository is necessarily text-heavy, I figured it might make sense to set up [`rumdl`](https://rumdl.dev/) for linting and formatting Markdown files.

I'm happy to hear any feedback on the configuration. If the changes are too invasive, we can also close this PR.


.github/CODEOWNERS
denialhaag, line 1: We are currently linking to this file, but it hasn't existed so far.

https://github.com/unitaryfoundation/jeff/blob/0573644037777823c2ef8a2a855fc24b65e17487/DEVELOPMENT.md#L201

For `rumdl` to be happy, we can either add this file or remove the link to it. 🤔

pyproject.toml
```diff
@@ -3,18 +3,41 @@ members = ["impl/py"]
...
+[tool.rumdl.per-file-ignores]
+# The RFCs deliberatively have a non-compliant format
+"rfcs/**" = ["MD025", "MD041", "MD053"]
+# The README starts with an image and may include HTML tags
+"README.md" = ["MD033", "MD041"]
+
+[tool.rumdl.MD013]
+line-length = 80
+reflow = true
+reflow-mode = "semantic-line-breaks"
```
denialhaag: This is arguably the most invasive change. There are other `reflow-mode`s: https://rumdl.dev/md013/#reflow-modes

.pre-commit-config.yaml
denialhaag, line 1: I moved `ruff check` before `ruff format` here, as this is the recommended order: https://github.com/astral-sh/ruff-pre-commit#using-ruff-with-pre-commit

.github/CODEOWNERS
aborgna-q: Yeah, I added with future release machinery in mind but we should probably remove it from the DEVELOPMENT.md for now.

examples/catalyst_simple/README.md
```diff
@@ -1,12 +1,13 @@
...
+including functions, numeric constants, quantum registers,
+and basic and parametrized gates as well as measurement.
 
-The Catalyst converter prototype can be found [here](https://github.com/PennyLaneAI/catalyst-jeff).
+The Catalyst converter prototype can be found at
+[PennyLaneAI/catalyst-jeff](https://github.com/PennyLaneAI/catalyst-jeff).
 
-
-### Catalyst MLIR
+## Catalyst MLIR
```
aborgna-q, line 10: Note that there are some render changes due to `rumdl` changing the header levels.

I'm not against it, but flaging it for attention.

pyproject.toml
aborgna-q: LGTM

aborgna-q (approved):
Nice, I didn't know this linter.

Should we add a CI check too?

burgholzer (approved):
LGTM 👍🏻 A CI check would probably be nice if we are not running pre-commit by default in CI

.github/CODEOWNERS
denialhaag: Makes sense! I've removed it again now. 👍

examples/catalyst_simple/README.md
denialhaag: Thanks for pointing that out! I went through the changes again and found some inconsistencies, which I've fixed in https://github.com/unitaryfoundation/jeff/pull/73/changes/62395f1e09002c1b9b079200e0b06b7934717de8.

denialhaag:
Thanks for the reviews!

I've added a CI check in https://github.com/unitaryfoundation/jeff/pull/73/commits/311a15f75d2c374f2b9c534e16b518cdd02c5dc2. Could someone change the branch detection accordingly? 🙂 

While at it, I've updated `dorny/paths-filter` in https://github.com/unitaryfoundation/jeff/pull/73/commits/c5dd83248217cfa5751656875f068842cf752a6b because `v3` was still using Node.js 20, which is deprecated.

README.md
```diff
@@ -7,40 +7,51 @@
     <img src="./docs/assets/JEFF_github_logo_inverted.png#gh-dark-mode-only" width="700px" onerror="this.style.display='none'" alt=""/>
 </p>
 
-`jeff` is a structured, extensible, and simple interchange format for quantum compilers. Designed
-to foster collaboration and accelerate research by providing a common language
+`jeff` is a structured, extensible,
+and simple interchange format for quantum compilers.
+Designed to foster collaboration
+and accelerate research by providing a common language
 for compilers to communicate.
```
dime10: It's a bit strange how short these lines are being made 🤔

burgholzer: It tries to break at semantic barriers 🤷🏼 
Definitely takes some time to get used to, but does not feel too bad.
In the end, it's one of the settings though and one can also change it to just break at the character limit without any heuristic.

We could also increase the line length from the default of 80 if we feel like it.

denialhaag: Yeah, this is a result of using `line-length = 80` and `reflow-mode = "semantic-line-breaks"`. Any line that is longer than 80 characters is broken at a semantic break (in this case, "and") and not as close as possible to the 80-character limit.

dime10: Oh I see, interesting. Definitely not what I'm used to, but if you prefer this I can live with it 😅

burgholzer: I have no particular personal preference here as long as we enforce _something_ 😌

denialhaag: @dime10, I just tried switching to the [`default` mode](https://rumdl.dev/md013/#default-mode) (see https://github.com/unitaryfoundation/jeff/pull/73/changes/47a12ddcbc7be836a7b0f3fa5018d4bd63259bc6), but it requires manually deleting all currently existing line breaks, which is why I gave up (see https://github.com/unitaryfoundation/jeff/pull/73/commits/79556d56de67f127dd5ee87a0980c9ab16bb51a1). Are you okay with sticking with semantic line breaks? While it might not be pretty, it does simplify maintenance because there is no room for interpretation.

dime10: I guess `normalize` might fix the issue with existing line breaks and will produce a consistent output?

denialhaag: It's almost impressive that I didn't see this myself, given that it's immediately visible when you click the link.

https://github.com/unitaryfoundation/jeff/pull/73/commits/62493e9dcb710962ac941a92b8c0010e8ec5c2b8 switches `reflow-mode` to `normalize`.

@aborgna-q, @burgholzer, and @dime10, let me know what you all think! If everyone is happy with this, I'd say we go ahead and merge. 😌

dime10: Nice looks good 👍

burgholzer (approved):
Let's get this in 👍🏻

PR #74, opened 2026-06-30
docs: Document that values must be defined before use

denialhaag:
This PR documents our decision to require values to be defined before use.


impl/cpp/src/capnp/jeff.capnp.c++
dime10, line 1: Weird that the bindings changed from updating the comments

dime10 (approved):
Thanks @denialhaag, looks good 👍

burgholzer (approved):
LGTM

impl/cpp/src/capnp/jeff.capnp.c++
aborgna-q: Yes, the generated files are deterministic, but any change to the `.capnp` changes them -.-

Issue #45, opened 2026-03-05
Add workflow for publishing `jeff-format` to PyPI


denialhaag:
As we released `v0.1.0` last week, it would make sense to also publish a first version of `jeff-format` on PyPI. Ideally, this should happen automatically for future releases via a workflow triggered on release.

See https://github.com/PennyLaneAI/jeff-mlir/pull/5#discussion_r2848587825 for an initial discussion on this.

denialhaag:
Closing this as completed by #90.

PR #75, opened 2026-07-01
docs: Prepare release of `jeff-v0.3.0`

denialhaag:
This PR prepares the release of `jeff-v0.3.0` by updating the changelog.


denialhaag:
If you all agree, I'd say we can release `jeff-v0.3.0` once #72 is merged. Would any of you prefer we hold off and change something else before a release?

While not a _must_, I'd like to pin a released version in https://github.com/PennyLaneAI/jeff-mlir/pull/23, so I'd prefer we release rather soon. 🙂

burgholzer (changes requested):
This generally looks good to me 👍🏼 

I was a bit surprised that this only changes the changelog and no other places. Then I found https://github.com/unitaryfoundation/jeff/blob/6ef694bf977c059252199e0fe15493737c59cf10/impl/py/pyproject.toml which we definitely forgot to update previously. Maybe it is worth checking for other places where a version number is off.

We should probably also start publishing the Python package at some point. Do we have the respective infrastructure set up already?
There is no direct need to do this as part of this release series; but I would assume it to be beneficial in the midterm.

denialhaag:
> I was a bit surprised that this only changes the changelog and no other places. Then I found https://github.com/unitaryfoundation/jeff/blob/6ef694bf977c059252199e0fe15493737c59cf10/impl/py/pyproject.toml which we definitely forgot to update previously. Maybe it is worth checking for other places where a version number is off.

Yeah, good point! We also have changelogs for [Python](https://github.com/unitaryfoundation/jeff/blob/main/impl/py/CHANGELOG.md) and [Rust](https://github.com/unitaryfoundation/jeff/blob/main/impl/rs/CHANGELOG.md), but I didn't update those because I figured that only makes sense if we actually release the package and the crate. I'm not entirely sure what the strategy is here. Does every GitHub release necessarily include a PyPI and a crates.io release? Are the versions necessarily aligned?
 
> We should probably also start publishing the Python package at some point. Do we have the respective infrastructure set up already? There is no direct need to do this as part of this release series; but I would assume it to be beneficial in the midterm.

[We do have documentation](https://github.com/unitaryfoundation/jeff/blob/main/DEVELOPMENT.md#python-package-release) on releasing the Python package, but I don't _think_ that `release-please` is actually set up. I created https://github.com/unitaryfoundation/jeff/issues/45 a while back to at least track this.

On a related note, I'm now relying on the Python package in https://github.com/PennyLaneAI/jeff-mlir/pull/23, but I'm using `uv` to install it from GitHub for now.

CHANGELOG.md
```diff
@@ -10,9 +10,26 @@ files.
...
+  by the new `while` operation ([#72])
+- A new `verifier` tool has been added to check the validity of a `jeff` module.
+  The `verifier` crate exports a `verify_file` function that loads and verifies
+  a `.jeff` file ([#68], [#72])
+
 ## [0.2.0] - 2026-04-14
 
-This version releases two breaking changes:
+This version releases two breaking changes. The full breakdown of the changes
+can be found below.
```
dime10: Usually one should not modify previous release notes, ideally they are immutable (except to fix a mistake in the notes themselves perhaps):
```suggestion
This version releases two breaking changes:
```

dime10 (approved):
Looks good thanks!

We can go ahead releasing the spec without worrying about the bindings. 
For the Python ones specifically they were never released because they never reached full coverage of the spec.

CHANGELOG.md
```diff
@@ -10,9 +10,26 @@ files.
...
+- The `doWhile` operation has been removed as its functionality is fully covered
+  by the new `while` operation ([#72])
+- A new `verifier` tool has been added to check the validity of a `jeff` module.
+  The `verifier` crate exports a `verify_file` function that loads and verifies
+  a `.jeff` file ([#68], [#72])
+
 ## [0.2.0] - 2026-04-14
 
-This version releases two breaking changes:
+This version releases two breaking changes. The full breakdown of the changes
```
dime10:
```suggestion
```

burgholzer (approved):
Except for the things David already mentioned, this LGTM 👍🏼

aborgna-q (approved):

PR #78, opened 2026-07-10
chore: Update Cap'n Proto to version 1.5.0

denialhaag:
This PR updates Cap'n Proto to version 1.5.0.


burgholzer (approved):
Probably deserves a changelog entry.
I am also surprised that nothing else has to be generated based on the updated version, but that's probably just me.

denialhaag:
> Probably deserves a changelog entry. 

Already done in https://github.com/unitaryfoundation/jeff/pull/75 with https://github.com/unitaryfoundation/jeff/pull/75/changes/9f50aa6e995b25b2cfd0483b3af62a9d23b67b8f.

> I am also surprised that nothing else has to be generated based on the updated version, but that's probably just me.

Yeah, I was also surprised. 🤔

aborgna-q (approved):
And the tests pass, so it seems not to be an issue with the language specific capnp depednencies 🤷

PR #89, opened 2026-08-03
chore(deps): Update the capnpc rust library

aborgna-q:
Update the capnproto library used on the rust crate, and update the generated bindings acordingly.


denialhaag (approved):
LGTM!

PR #91, opened 2026-08-07
ci: Enable pr-title CI check

aborgna-q:
Now that we are automatically generating release changelogs from the pr titles, we should make sure that the titles follow [semantic versioning](https://semver.org/) so they can be detected by the bots.

There was a disabled workflow before, that required a bot github token to work. This PR replaces it with a simpler definition that doesn't try to post in the thread.
Ignore the red check from the old one here.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,6 +1,7 @@
-name: Check Conventional Commits format
+name: PR title check 📄
+
 on:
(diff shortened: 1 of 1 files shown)


denialhaag (approved):
LGTM! 😎 

We probably have to adapt the branch protection. 🤔

aborgna-q:
Yes, we should add this check as required

PR #96, opened 2026-08-14
feat: Add `jeff` programs for structured benchmarks

denialhaag:
🤖 *AI text below* 🤖

Adds binary `.jeff` programs for all structured benchmark families currently available through the MQT Core benchmark library. Each scalable family has instances for `n = 3`, `n = 5`, and `n = 7`; teleportation remains fixed-size. The classical-input and quantum-input QFT adders are tracked separately.

Replaces the OpenQASM reference programs with representative QC MLIR views for `n = 3`. These views are provided only for inspection. Benchmarks should use the binary `.jeff` programs.

Adds a Python recipe for generating other sizes with MQT Core v4. Further benchmarks will be added on a rolling basis.


benchmarks/_recipes/structured/generate.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+
+# /// script
+# requires-python = ">=3.11"
+# dependencies = ["mqt-core>=4"]
```
aborgna-q: Re-running the script produces different `.jeff`s than what's checked in.

Can we pin a specific `mqt-core` here?

benchmarks/structured/README.md
```diff
@@ -71,24 +91,33 @@ with the structured control-flow primitives they employ.
...
+| Measurement-based quantum computation | 🟦                       | ❌                        | 🟦                     | ✔️                       | ❌                                          | ✔️                                  | ❌                       | 🟦          | [Wikipedia](https://en.wikipedia.org/wiki/One-way_quantum_computer)                                 | ❌             | 🟦        |
+
+## Generate Other Sizes
+
+The programs were generated with MQT Core. Run the Python recipe from the
+repository root with `uv`:
+
+```console
+uv run benchmarks/_recipes/structured/generate.py 9
+```
```
aborgna-q: Add this as a recipe to the root `justfile`.

aborgna-q (approved):

benchmarks/_recipes/structured/generate.py
aborgna-q: Looks like the `jeff`s were generated with `mqt-cc @ 3.3.4`, did the script get updated afterwards?

denialhaag: Thanks for catching that! I pinned `~=4.0.0` and regenerated the benchmarks in https://github.com/unitaryfoundation/jeff/pull/96/commits/61e50dabb8ed4b4af6df3846d0897ead478aeb20. 🙂

benchmarks/structured/README.md
denialhaag: Done in https://github.com/unitaryfoundation/jeff/pull/96/changes/e42f85ba95426c3f7c516b6ef59f4ecf6f82b72f.

benchmarks/_recipes/structured/generate.py
denialhaag: (I'm all but certain there were no significant changes to `mqt-cc` since the last time I generated the benchmarks. `mqt-cc @ 3.3.4` only refers to a development version.)

aborgna-q: Can you pin `==4.0.0` instead? So an update doesn't break reproducibility.

justfile
```diff
@@ -61,3 +61,7 @@ update-capnp:
     patch -p0 -R < impl/capnp/cpp_namespace.patch
     # Re-encode the test examples
     ./examples/encode_examples.sh
+
+# Generate benchmark programs.
```
aborgna-q, line 65: Explain the behaviour of `*N*` here.

`just` doesn't yet support long docs, but we can leave it as a comment.

See [casey/just#2684](https://www.github.com/casey/just/issues/2684#issuecomment-5315660404).

```suggestion
# Generate benchmark programs.
#
# The positional arguments select values of `n >= 3`. Without arguments, the
# recipe recreates the checked-in programs for `n = 3`, `n = 5`, and `n = 7`. The
# benchmark READMEs explain how each family interprets `n`.
[doc('Generate benchmark programs.')]
```

benchmarks/_recipes/structured/generate.py
denialhaag: Makes sense! Done in https://github.com/unitaryfoundation/jeff/pull/96/commits/4c7e1bf5ba72e38ed2a5d9a4d7cee62c73adda7e.

benchmarks/structured/README.md
```diff
@@ -7,40 +7,60 @@ techniques.
...
+
+> [!NOTE]
+> Additional benchmarks will be added on a rolling basis.
+
+The QC MLIR files are representative, human-readable views of the generated
+programs for `n = 3`. QC is one of the MLIR dialects in the
+[MQT Compiler Collection (`mqt-cc`)](https://mqt.readthedocs.io/projects/core/en/latest/mlir/index.html),
+which is part of [MQT Core](https://github.com/munich-quantum-toolkit/core).
+These files are provided only for inspection. Use the binary `.jeff` files for
+benchmarks. Teleportation has a fixed size and therefore has no `n` suffix.
```
DRovara: I think we need a linebreak before the last line here, otherwise it looks a bit weird.

DRovara (approved):
One tiny stylistic recommendation, otherwise it looks good to me.

Thanks a lot @denialhaag for pushing this forward!

benchmarks/structured/README.md
denialhaag: Thanks for pointing that out! Fixed in https://github.com/unitaryfoundation/jeff/pull/96/commits/9103a2a7fc2d914ded5b312ae53e14ebf67d8dc5.

burgholzer (approved):
This looks pretty good to me 👍🏼 Thanks @denialhaag 🙏🏼

denialhaag:
Thanks for all the feedback, everybody! I'll merge this now so that the benchmarks are available when the paper comes out. :shipit:

PR #101, opened 2026-08-22
feat: Add more arithmetic operations

### LINKED ISSUE #97: Add support for converting between `int` and `float`
🤖 *AI text below* 🤖

The classical operations (https://github.com/unitaryfoundation/jeff/blob/main/docs/spec.md#classical-operations-intop-floatop-etc) offer no way to convert between integers and floats.

This makes it impossible to derive a float from an integer that is only known at runtime. A concrete case is a rotation angle that depends on a loop index, such as `p(pi/2**(j-i))` in a QFT. Without a conversion, the angle cannot be computed from the induction variable and has to be carried through the loop and rescaled on each iteration instead.

`ceil` and `floor` are defined, but they take and return floats, so they cannot produce an integer either.

### LINKED ISSUE #98: Add support for dividing `float`s
🤖 *AI text below* 🤖

`FloatOp` defines `add`, `sub`, `mul`, and `pow`, but no division (https://github.com/unitaryfoundation/jeff/blob/main/docs/spec.md#floatop--floatarrayop). `IntOp` has `divS` and `divU`, so the gap is specific to floats.

Dividing by a value that is only known at runtime therefore cannot be expressed. Multiplying by a reciprocal works only when the divisor is a constant.

denialhaag:
This PR adds a range of new arithmetic operations. Most of these came up while working on generating `jeff` versions of the [structured programs](https://github.com/unitaryfoundation/jeff/tree/main/benchmarks/structured). They are largely inspired by [MLIR's `arith` dialect](https://mlir.llvm.org/docs/Dialects/ArithOps/).

Fixes #97
Fixes #98


impl/capnp/jeff.capnp
```diff
@@ -930,6 +931,62 @@ struct IntOp {
...
+        select @29 :Void;
+        # Select between two integers based on a condition mask.
+        #
+        # Inputs:
+        # - `int(1)`: Condition mask.
+        # - `int(N)`: Value selected when condition is true.
+        # - `int(N)`: Value selected when condition is false.
+        #
+        # Outputs:
+        # - `int(N)`: Selected integer output.
```
denialhaag, line 944: This is arguably already covered by `ScfOp.switch`, but I think it's still a nice addition.

impl/capnp/jeff.capnp
```diff
@@ -930,6 +931,62 @@ struct IntOp {
...
+        # - `float(M)`: The converted float.
+
+        toFloatU @34 :Void;
+        # Convert an unsigned integer to a float.
+        #
+        # Inputs:
+        # - `uint(N)`: The unsigned integer operand.
+        #
+        # Outputs:
+        # - `float(M)`: The converted float.
```
denialhaag, line 989: I wasn't really sure where to put these and their `toSInt` and `toUInt` counterparts. Do we want to define a new `ConversionOp` type, or are they fine where they are? 🤔

denialhaag (commented):
Some context/open questions:

burgholzer (approved):
Generally looks reasonable 👍🏻
I am not sure whether a dedicated conversion module would be better; kind of feels fine as it is.
As for the select operation, I also think it makes sense to add, as it is much simpler than its SCF counterpart and can more easily be mapped to arith.select in MLIR without complicated analysis.

tools/verifier/src/passes/type_checks.rs
```diff
@@ -435,6 +549,64 @@ fn check_float_op(
...
+            );
+            expect_output(
+                outputs,
+                0,
+                |ty| is_int(ty, None),
+                "float-to-int conversion",
+                errors,
+            );
+        }
         _ => {}
```
aborgna-q, line 615: Missing `FloatOp::Div`

tools/verifier/src/passes/type_checks.rs
```diff
@@ -363,6 +443,40 @@ fn check_int_op(
             check_uniform_int(inputs, &[], "int comparison", errors);
             expect_output(outputs, 0, |ty| is_int(ty, 1), "int comparison", errors);
         }
+        IntOp::Select => {
+            check_arity(inputs, 3, outputs, 1, "int select", errors);
+            expect_input(inputs, 0, |ty| is_int(ty, 1), "int select", errors);
+            check_uniform_int(&inputs[1..], outputs, "int select", errors);
```
aborgna-q:
```suggestion           check_uniform_int(&inputs.get(1..).unwrap_or_default(), outputs, "int select", errors);
```
Otherwise this will panic if the inputs are empty

tools/verifier/src/passes/type_checks.rs
```diff
@@ -435,6 +549,64 @@ fn check_float_op(
             check_uniform_float(inputs, &[], "float predicate", errors);
             expect_output(outputs, 0, |ty| is_int(ty, 1), "float predicate", errors);
         }
+        FloatOp::Select => {
+            check_arity(inputs, 3, outputs, 1, "float select", errors);
+            expect_input(inputs, 0, |ty| is_int(ty, 1), "float select", errors);
+            check_uniform_float(&inputs[1..], outputs, "float select", errors);
```
aborgna-q:
```suggestion
            check_uniform_float(&inputs.get(1..).unwrap_or_default(), outputs, "float select", errors);
```

impl/capnp/jeff.capnp
```diff
@@ -1312,6 +1379,53 @@ struct FloatOp {
...
+        # - `float(M)`: The extended higher-precision float value (where M > N).
+
+        trunc @35 :Void;
+        # Truncate floating-point precision to a narrower float type.
+        #
+        # Inputs:
+        # - `float(N)`: The higher-precision float operand.
+        #
+        # Outputs:
+        # - `float(M)`: The truncated lower-precision float value (where M < N).toSInt @38 :Void;
```
aborgna-q:
```suggestion
        # - `float(M)`: The truncated lower-precision float value (where M < N)
```

tools/verifier/src/passes/type_checks.rs
denialhaag: Added in https://github.com/unitaryfoundation/jeff/pull/101/commits/a5d12f50a60924e146c9ea77c54eb2733c2e5884.

denialhaag:
Thanks a lot for the feedback, @aborgna-q! Everything should be fixed now! 🙂

denialhaag:
I'll merge this now to unblock https://github.com/unitaryfoundation/jeff-mlir/pull/44. We can easily follow up on this if something else needs fixing. I would like to create a release later this week once #96 is ready and merged. 😌

PR #109, opened 2026-09-14
ci: Fix automatic PyPi publication with trusted publishing

aborgna-q:
Copies the package publishing workflow from `python-release.yml` to `release-please.yml`.

We were calling the former via a `workflow_call` before, but PyPi Trusted Publishing doesn't support this. This PR just copies the publishing code to `release-please.yml` instead (and leaves the other workflow for manual releases / retries).

Error message:

```
 The claims in this token suggest that the calling workflow is a reusable workflow.
  
  In particular, this action was initiated by:
  
      unitaryfoundation/jeff/.github/workflows/python-release.yml@refs/heads/main
  
  Whereas its parent workflow is:
  
      unitaryfoundation/jeff/.github/workflows/release-please.yml@refs/heads/main
  
  Reusable workflows are **not currently supported** by PyPI's Trusted Publishing
  functionality, and are subject to breakage. Users are **strongly encouraged**
  to avoid using reusable workflows for Trusted Publishing until support
  becomes official. Please, do not report bugs if this breaks.
  
  For more information, see:
  
  * https://docs.pypi.org/trusted-publishers/troubleshooting/#reusable-workflows-on-github
  * https://github.com/pypa/gh-action-pypi-publish/issues/166 — subscribe to
    this issue to watch the progress and learn when reusable workflows become
    supported officially
 ```
 https://github.com/unitaryfoundation/jeff/actions/runs/34587907204/job/103227117525#step:10:147


denialhaag (approved):
Thanks a lot for fixing this! :shipit:

PR #112, opened 2026-09-15
fix(py): read the subkind of an op group that has no union

### LINKED ISSUE #111: jeff-py: reading back a func.funcCall op raises "Attempted to call which on a non-union type"
`JeffOp.subkind` calls `.which` on the instruction's kind group, but `FuncOp` is the only op struct without a union, so reading any `func.funcCall` op back from an encoding raises `KjException: Attempted to call which on a non-union type`. `instruction_data` goes through `subkind`, so a call's function index is unreachable too. Encoding works; only reading breaks.

**Example**:
```python
import jeff as jf

call = jf.JeffOp("func", "funcCall", [], [], 1)
main = jf.FunctionDef(name="main", body=jf.JeffRegion(sources=[], targets=[], operations=[call]))
callee = jf.FunctionDef(name="callee", body=jf.JeffRegion(sources=[], targets=[], operations=[]))
module = jf.JeffModule([main, callee])
module.refresh()

op = jf.JeffModule.from_encoding(module._raw_data).functions[0].body.operations[0]
print(op.kind)             # func
print(op.subkind)          # KjException: Attempted to call which on a non-union type
print(op.instruction_data) # never reached
```

nathanieltornow:
A group without union fields now reports its only field, so `func.funcCall` ops read back through `subkind` and `instruction_data`.

Closes #111


impl/py/tests/test_reader.py
```diff
@@ -1,2 +1,25 @@
+import jeff as jf
```
denialhaag: I'm not sure whether we really want to establish `jf`. 🤔

impl/py/tests/test_reader.py
```diff
@@ -1,2 +1,25 @@
...
+
+
 def test_hello_world() -> None:
     assert 2 + 2 != "🐟"
```
denialhaag: This can probably be removed at this point.

denialhaag (commented):
Thanks a lot for the fix, @nathanieltornow! 🙂

This looks pretty good to me. I have two minor comments below. I'll request another review for final approval.

nathanieltornow:
Thanks @denialhaag, I've implemented your suggestions :)

denialhaag (approved):
Looks good from my side now! 🙂

burgholzer (approved):
LGTM! Thanks for the fix!

